PR #1, opened 2018-03-14
Controller side implementation of replication

payes:


src/replication.c
```diff
@@ -0,0 +1,1067 @@
...
+						}
+						rcommq_ptr->completed = 1;
+					}
+					MTX_UNLOCK(&spec->rcommonq_mtx);
+
+					MTX_LOCK(&spec->luworker_rmutex[luworker_id]);
+					pthread_cond_signal(&spec->luworker_rcond[luworker_id]);
+					MTX_UNLOCK(&spec->luworker_rmutex[luworker_id]);
+
+				} else if (rcommq_ptr->acks_recvd == spec->replica_count) {
```
vishnuitta: maintain the count to which IOs are sent, rather than looking at replica_count which might change

src/replication.c
```diff
@@ -0,0 +1,1067 @@
...
+		if(replica_count != spec->replica_count) {
+			replica_count = update_replica_list(epfd, spec, replica_count);
+		}
+		//Wait for events on all iofds
+		event_count = epoll_wait(epfd, events, MAXEVENTS, -1);
+		for(i=0; i< event_count; i++) {
+			//Remove the replica from queue
+			if ((events[i].events & EPOLLERR) ||
+					(events[i].events & EPOLLHUP) ||
+					(!(events[i].events & EPOLLIN))) {
```
vishnuitta: EINTR should not be treated as error

vishnuitta: same with other places also..
actually, I have one doubt also.. do we get events only for those which are set in event.events? :)

src/replication.c
```diff
@@ -0,0 +1,1067 @@
...
+		event.data.fd = mgmt_fd;
+		event.events = EPOLLIN | EPOLLET;
+		rc = epoll_ctl(epfd, EPOLL_CTL_ADD, mgmt_fd, &event);
+		if(rc == -1) {
+			REPLICA_ERRLOG("epoll_ctl() failed, errno:%d", errno);
+			return -1;
+		}
+		MTX_LOCK(&specq_mtx);
+		TAILQ_FOREACH(spec, &spec_q, spec_next) {
+			data_len += snprintf(buf, BUFSIZE, "%s", "vol1") + 1;
```
vishnuitta: this need to be made configurable, and its at multiple places.

src/replication.c
```diff
@@ -0,0 +1,1067 @@
...
+	build_replica_mgmt_hdr();
+	iovec[0].iov_base = rmgmtio;
+	nbytes = iovec[0].iov_len = sizeof(zvol_io_hdr_t);
+	iovec[1].iov_base = buf;
+	iovec[1].iov_len = rmgmtio->len;
+	nbytes += rmgmtio->len;
+	iovcnt = 2;
+	while (nbytes) {
+		rc = writev(fd, iovec, 2);//Review iovec in this line
+		if (rc < 0) {
```
vishnuitta: need to handle EAGAIN return value, and at multiple other places where writev is used

src/replication.c
```diff
@@ -0,0 +1,1067 @@
...
+	return 0;
+}
+
+int
+accept_mgmt_conns(int epfd, int sfd) {
+	struct epoll_event event;
+	int rc, rcount=0;
+	spec_t *spec;
+	int mgmtfd[MAXREPLICA];
+	char *buf = malloc(BUFSIZE);
```
vishnuitta: this buf is not freed

src/replication.c
vishnuitta: EPOLLOUT event of fd seems to be not part of epoll. This should be good.

src/replication.c
```diff
@@ -0,0 +1,1067 @@
...
+	return 0;
+}
+
+int64_t
+read_data(int fd, uint8_t *data, uint64_t len) {
+	int64_t rc;
+	uint64_t nbytes = 0;
+	while((rc = read(fd, data + nbytes, len - nbytes))) {
+		if(rc < 0) {
+			if(nbytes > 0 && errno == EAGAIN) {
```
vishnuitta: should be nbytes >= 0?

src/replication.c
```diff
@@ -0,0 +1,1067 @@
...
+			} else if (events[i].data.fd == sfd) {
+				//Accept management connections from replicas and add the replicas to replica queue
+				mgmtfd = accept_mgmt_conns(epfd, sfd);
+				continue;
+			} else {
+				//Negotiate operating parameters with replica and add the replica to spec->rq
+					count = read_data(events[i].data.fd, (uint8_t *)mgmt_ack, sizeof(zvol_io_hdr_t));
+					if (count == -1 && errno == EAGAIN) {
+						//TODO HANDLE DATA for next iteration
+						break;
```
vishnuitta: this will break entire 'for' loop.. we should do continue

src/replication.c
```diff
@@ -0,0 +1,1067 @@
...
+					}
+					check_read_error();
+
+					if(mgmt_ack->len) {
+						mgmt_ack_data = malloc(sizeof(mgmt_ack_data_t));
+						count = read_data(events[i].data.fd, (uint8_t *)mgmt_ack_data, mgmt_ack->len);
```
vishnuitta: we need to make sure mgmt_ack->len is sizeof(mgmt_ack_data_t)

src/replication.c
```diff
@@ -0,0 +1,1067 @@
...
+zvol_handshake(char *volname, char *replicaip, int replica_port) {
+	int rc, iofd;
+	spec_t *spec;
+	REPLICA_LOG("VOLNAME = %s\n", volname);
+	MTX_LOCK(&specq_mtx);
+	TAILQ_FOREACH(spec, &spec_q, spec_next) {
+		if(strcmp(volname, volname) == 0)  {
+			if((iofd = cstor_ops.conn_connect(replicaip, replica_port)) < 0) {
+				REPLICA_ERRLOG("conn_connect() failed errno:%d\n", errno);
+				exit(EXIT_FAILURE);
```
vishnuitta: exit's need to be avoided

payes:
A new branch has been created to review these changes. Hence closing this PR.

src/replication.c
payes: done in latest commit

PR #3, opened 2018-04-02
[cstor#120] Pick replication options from config file Issue

payes:
istgt code was using hard coded volume name as 'vol1'. This needs to be picked from configuration file.
Replication and consistency factor also need to be picked from istgt.conf file.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -777,12 +777,9 @@ create_replica_entry(spec_t *spec, int iofd, char *replicaip, int replica_port) 
 	rio->opcode = ZVOL_OPCODE_HANDSHAKE;
 	rio->io_seq = 0;
 	rio->offset = 0;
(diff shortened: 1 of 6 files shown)


src/replication.c
```diff
@@ -864,7 +864,7 @@ accept_mgmt_conns(int epfd, int sfd) {
 		}
 		MTX_LOCK(&specq_mtx);
 		TAILQ_FOREACH(spec, &spec_q, spec_next) {
-			data_len += snprintf(buf, BUFSIZE, "%s", "vol1") + 1;
+			data_len += snprintf(buf, BUFSIZE, "%s", spec->lu->volname) + 1;
```
vishnuitta: there should be only one spec in spec_q.. do a 'break' after this..

PR #5, opened 2018-04-03
[cstor#121] Halt IOs till required number of replicas are connected to target

payes:
Add degraded and healthy modes for replica.
Start serving IOs only when consistency factor has been met.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -192,6 +192,9 @@ dequeue_common_sendq:
 			MTX_UNLOCK(&replica->r_mtx);
 			if((rcmd->opcode == ZVOL_OPCODE_READ) && read_cmd_sent) {
 				break;
+			} else if(rcmd->opcode == ZVOL_OPCODE_WRITE) {
+				//TODO This needs to be added in the starting
+				cmd->copies_sent++;
 			}
 		}
 	}
@@ -361,7 +364,7 @@ remove_replica_from_list(spec_t *spec, int iofd) {
 					MTX_LOCK(&spec->luworker_rmutex[luworker_id]);
 					pthread_cond_signal(&spec->luworker_rcond[luworker_id]); 
 					MTX_UNLOCK(&spec->luworker_rmutex[luworker_id]);
-				}else if(rcommq_ptr->acks_recvd == spec->replica_count - 1) {
+				}else if(rcommq_ptr->acks_recvd == rcommq_ptr->copies_sent - 1) {
 					rcommq_ptr->status = -1;
 					MTX_LOCK(&spec->luworker_rmutex[luworker_id]);
 					pthread_cond_signal(&spec->luworker_rcond[luworker_id]); 
@@ -380,7 +383,7 @@ remove_replica_from_list(spec_t *spec, int iofd) {
 					MTX_LOCK(&spec->luworker_rmutex[luworker_id]);
 					pthread_cond_signal(&spec->luworker_rcond[luworker_id]); 
 					MTX_UNLOCK(&spec->luworker_rmutex[luworker_id]);
-				}else if(rcommq_ptr->acks_recvd == spec->replica_count - 1) {
+				}else if(rcommq_ptr->acks_recvd == rcommq_ptr->copies_sent - 1) {
 					rcommq_ptr->status = -1;
 					MTX_LOCK(&spec->luworker_rmutex[luworker_id]);
 					pthread_cond_signal(&spec->luworker_rcond[luworker_id]); 
@@ -394,7 +397,9 @@ remove_replica_from_list(spec_t *spec, int iofd) {
 			pthread_cond_signal(&replica->r_cond);
 			MTX_UNLOCK(&replica->r_mtx);
 			spec->replica_count--;
-			if(spec->replica_count == 0){
+			if((spec->healthy_rcount < spec->consistency_factor) &&
+				(MAX(spec->replication_factor - spec->consistency_factor + 1, spec->consistency_factor) <
+					 spec->healthy_rcount  + spec->degraded_rcount)) {
 				spec->ready = false;
 			}
 			TAILQ_REMOVE(&spec->rq, replica, r_next);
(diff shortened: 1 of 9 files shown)


src/replication.c
```diff
@@ -192,6 +192,9 @@ replicator(void *arg) {
 			MTX_UNLOCK(&replica->r_mtx);
 			if((rcmd->opcode == ZVOL_OPCODE_READ) && read_cmd_sent) {
 				break;
+			} else if(rcmd->opcode == ZVOL_OPCODE_WRITE) {
+				//TODO This needs to be added in the starting
+				cmd->copies_sent++;
```
vishnuitta: we can do this for read IOs also.. and do this right after inserting into sendq of replica

src/replication.c
```diff
@@ -394,7 +397,9 @@ remove_replica_from_list(spec_t *spec, int iofd) {
 			pthread_cond_signal(&replica->r_cond);
 			MTX_UNLOCK(&replica->r_mtx);
 			spec->replica_count--;
-			if(spec->replica_count == 0){
+			if((spec->healthy_rcount < spec->consistency_factor) &&
```
vishnuitta: condition to allow IOs should be: (n is replication factor, m is consistency factor)
- MAX(n-m+1, m) number of degraded replicas (or)
- 'm' number of connected replicas with atleast one healthy replica

vishnuitta: first condition is edited

src/replication.c
```diff
@@ -394,7 +397,9 @@ remove_replica_from_list(spec_t *spec, int iofd) {
 			pthread_cond_signal(&replica->r_cond);
 			MTX_UNLOCK(&replica->r_mtx);
 			spec->replica_count--;
-			if(spec->replica_count == 0){
```
vishnuitta: Add a TODO statement to decrement health/degraded count based on the state of replica when it got disconnected

src/replication.c
```diff
@@ -361,7 +364,7 @@ remove_replica_from_list(spec_t *spec, int iofd) {
 					MTX_LOCK(&spec->luworker_rmutex[luworker_id]);
 					pthread_cond_signal(&spec->luworker_rcond[luworker_id]); 
 					MTX_UNLOCK(&spec->luworker_rmutex[luworker_id]);
-				}else if(rcommq_ptr->acks_recvd == spec->replica_count - 1) {
+				}else if(rcommq_ptr->acks_recvd == rcommq_ptr->copies_sent - 1) {
```
vishnuitta: we need to do this only if acks_recvd is not met consistency_factor.. same case for below one also

vishnuitta: check whether we need to unset bit in 'bitset' while removing from waitq / blockedq

src/replication.c
```diff
@@ -605,12 +610,12 @@ handle_write_resp(spec_t *spec, replica_t *replica, zvol_io_hdr_t *io_rsp)
...
+				if (rcommq_ptr->acks_recvd == spec->consistency_factor) {
 					rcommq_ptr->status = 1;
 					//Transfer cmd from waitq to pendingq and wakeup luworker
 					MTX_LOCK(&spec->rcommonq_mtx);
 					TAILQ_REMOVE(&spec->rcommon_waitq, rcommq_ptr, wait_cmd_next);
-					if (rcommq_ptr->acks_recvd != spec->replica_count) {
+					if (rcommq_ptr->acks_recvd != rcommq_ptr->copies_sent) {
 						TAILQ_INSERT_TAIL(&spec->rcommon_pendingq, rcommq_ptr, pending_cmd_next);
 					} else {
 						/* we can leave this free to lbwrite */
```
vishnuitta: remove this comment, and also below one that says 'else is not correct..'

src/replication.c
```diff
@@ -344,64 +349,80 @@ update_replica_list(int epfd, spec_t *spec, int replica_count) {
...
+	rcommq_ptr->bitset &= ~(1 << replica->id);
+	//TODO Add check for acks_received in read also, same as write
+	if(rep_cmd->opcode == ZVOL_OPCODE_READ) {
+		rcommq_ptr->status = -1;
+		rcommq_ptr->state = CMD_EXECUTION_COMPLETED;
+		signal_luworker();
+	} else if(rep_cmd->opcode == ZVOL_OPCODE_WRITE) {
+		if(rcommq_ptr->acks_recvd + rcommq_ptr->ios_aborted
+				== rcommq_ptr->copies_sent - 1) {
+			if (rcommq_ptr->acks_recvd < spec->consistency_factor) {
```
vishnuitta: for sake of consistency, increase ios_aborted here also

src/replication.c
```diff
@@ -344,64 +349,80 @@ update_replica_list(int epfd, spec_t *spec, int replica_count) {
...
+	} else {
+		spec->ready = false;
+	}
+}
+
+void clear_replica_cmd(spec_t *spec, replica_t *replica, rcmd_t *rep_cmd) {
+	int i;
+	rcommon_cmd_t *rcommq_ptr = rep_cmd->rcommq_ptr;
+	int luworker_id = rcommq_ptr->luworker_id;
+	rcommq_ptr->bitset &= ~(1 << replica->id);
```
vishnuitta: lock protection is required in using rcommon_cmd_t. created an issue #136 to make rcommon_cmd_t as thread safe.

src/replication.c
vishnuitta: Lets put a log here that data is committed to few repilcas, but, less than consistency factor, and errored out to client.

vishnuitta: in this case also, do we need to free iov_base, and remove from pendingq/waitq..?

src/replication.c
```diff
@@ -580,6 +604,52 @@ handle_read_resp(spec_t *spec, replica_t *replica, zvol_io_hdr_t *io_rsp, void *
...
+		} \
+		rcommq_ptr->state = CMD_EXECUTION_COMPLETED; \
+	} \
+	MTX_UNLOCK(&spec->rcommonq_mtx); \
+	signal_luworker(); \
+}
+
+#define handle_all_resp_recvd() { \
+	MTX_LOCK(&spec->rcommonq_mtx); \
+	TAILQ_REMOVE(&spec->rcommon_pendingq, rcommq_ptr, pending_cmd_next); \
```
vishnuitta: Here, we come if acks_recv < CF or acks_recv > CF. For second case, this code looks fine. There may be case where all resp recvd, but, consistency factor is not reached. So, we need to check the state of rcommand_t to remove from pendingq/waitq. we need to wakeup luworker in case of first one.

PR #9, opened 2018-04-09
[cstor#123] maintain epoll with fds of connected replicas as and when required in istgt

payes:
Remove dependency of updating epoll on epoll_wait.
Store epoll fd in spec, and modify epoll as and when required i.e. at the time of add or removing replica.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -265,29 +265,18 @@ write_to_socket:
 void *
 replica_sender(void *arg) {
 	replica_t *replica = (replica_t *)arg;
-	int epfd, rc;
+	int rc;
 	rcommon_cmd_t *cmd = NULL;
 	rcmd_t *rcmd = NULL;
-	struct epoll_event event;
 	rcommon_cmd_t *rcommq_ptr;//REMOVE
-	epfd = epoll_create1(0);
-
-	event.data.fd = replica->iofd;
-	event.events = EPOLLOUT;
-	rc = epoll_ctl(epfd, EPOLL_CTL_ADD, replica->iofd, &event);
-	if (rc == -1) {
-		REPLICA_LOG("ERROR\n");
-		exit(1);
-	}
 	while(1) {
 		MTX_LOCK(&replica->r_mtx);
 dequeue_rsendq:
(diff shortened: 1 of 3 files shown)


src/replication.c
```diff
@@ -808,6 +761,27 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd, char *replicaip
 	free(rio);
 	rio = NULL;
 	update_volstate(spec);
+	struct epoll_event event;
```
vishnuitta: Even though not related to this code change, please move addition of replica to spec->rq and update_volstate towards end of the function.
So, first all the housekeeping work like creating threads etc, then, adding replica to spec->rq and then update_volstate.

src/replication.c
```diff
@@ -685,10 +648,6 @@ replica_receiver(void *arg) {
 			}
 			MTX_UNLOCK(&spec->rq_mtx);
 		}
-		//Check if number of available replicas have changed
```
vishnuitta: replica cleanup need to happen in case of any errors while reading in this thread as part of cstor#143 bug.. we don't need this as part of this PR

src/replication.c
```diff
@@ -399,13 +355,15 @@ void clear_replica_cmd(spec_t *spec, replica_t *replica, rcmd_t *rep_cmd) {
...
-	int ios_aborted = 0;
+	int ios_aborted = 0, rc;
 	rcmd_t *rep_cmd = NULL;
+	struct epoll_event event;
 	MTX_LOCK(&spec->rq_mtx);
 	TAILQ_FOREACH(replica, &spec->rq, r_next) {
 		if(iofd == replica->iofd) {
 			REPLICA_LOG("REMOVE REPLICA FROM LIST\n");
 			MTX_LOCK(&replica->r_mtx);
+			replica->state = REPLICA_ERRORED;
```
vishnuitta: here also, order need to be maintained.
First, update_volstate after updating the states and healthy/degraded replicas counters (updating counters can be part of cstor#137), then, remove from linked list and then signaling to stop threads and then closing fds.

vishnuitta:
@payes please add more details to description of PR.

PR #10, opened 2018-04-11
[cstor-124]Fix issues with replica ids and replica cleanup

payes:
In istgt, replica's id is set to spec's (spec is data structure for a LUN specification) 'replica_count', where, 'replica_count' is the number of connected replicas.
So, if a replica gets disconnected and a new replica gets connected, there are chances that two replicas can have same replica id.
replica id is used to find the replicas on which a given IO is waiting, based on which new IOs will be added to blocked queue or command queue.
replica->id need to be made unique across replicas and almost sequential by filling out gap if any during addition of replica.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -15,7 +15,6 @@
 #include "istgt_crc32c.h"
 #include "istgt_misc.h"
 #include "ring_mempool.h"
-
 cstor_conn_ops_t cstor_ops = {
 	.conn_listen = replication_listen,
 	.conn_connect = replication_connect,
@@ -36,6 +35,7 @@ extern size_t rcommon_cmd_mempool_count;
 	rio->offset = cmd->offset;\
 	rio->len    = cmd->data_len;\
 }
+
 #define build_replica_mgmt_hdr() \
 {\
 	rmgmtio = (zvol_io_hdr_t *)malloc(sizeof(zvol_io_hdr_t));\
@@ -88,6 +88,21 @@ extern size_t rcommon_cmd_mempool_count;
 	MTX_UNLOCK(&spec->luworker_rmutex[luworker_id]); \
 }
 
+int allocate_replica_id(spec_t *spec) {
+	int replica_id = ffs(~spec->replica_ids)-1;
+	if(replica_id == -1) {
+		REPLICA_ERRLOG("Replica limit reached\n");
+		return -1;
+	}
+	spec->replica_ids |= (1 << replica_id);
+	return replica_id;
+}
+
+void release_replica_id(spec_t *spec, int replica_id) {
+	spec->replica_ids &= ~(1 << replica_id);
+}
+
+
 int
 send_mgmtio(int fd, zvol_op_code_t mgmt_opcode, void *buf, size_t data_len) {
 	zvol_io_hdr_t *rmgmtio = NULL;
@@ -128,6 +143,20 @@ send_mgmtio(int fd, zvol_op_code_t mgmt_opcode, void *buf, size_t data_len) {
 	return 0;
 }
 
+replica_t *get_next_replica(spec_t *spec, replica_t *replica) {
+	MTX_LOCK(&spec->rq_mtx);
+	if(replica == NULL) {
+		replica = TAILQ_FIRST(&spec->rq);
+	} else {
+		replica->refcount--;
+		replica = TAILQ_NEXT(replica, r_next);
+	}
+	if(replica != NULL)
+		replica->refcount++;
+	MTX_UNLOCK(&spec->rq_mtx);
+	return replica;
+}
+
 //Picks cmds from send queue and sends them to replica
 void *
 replicator(void *arg) {
@@ -155,7 +184,7 @@ dequeue_common_sendq:
 		cmd->bitset = 0;
 		TAILQ_REMOVE(&spec->rcommon_sendq, cmd, send_cmd_next);
 		cmd->state = CMD_ENQUEUED_TO_WAITQ; \
-		TAILQ_INSERT_TAIL(&spec->rcommon_waitq, cmd, wait_cmd_next);
+			     TAILQ_INSERT_TAIL(&spec->rcommon_waitq, cmd, wait_cmd_next);
 		//Check for blockage in rcommon_pendingq and set corresponding blocked replica bits
 		bitset = 0;
 		TAILQ_FOREACH(pending_cmd, &spec->rcommon_pendingq, pending_cmd_next) {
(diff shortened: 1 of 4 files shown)


src/replication.c
```diff
@@ -265,6 +281,7 @@ sendio(int epfd, int fd, rcommon_cmd_t *cmd, rcmd_t *rcmd) {
 void *
 replica_sender(void *arg) {
 	replica_t *replica = (replica_t *)arg;
+	__sync_fetch_and_add(&replica->refcount, 1);
```
vishnuitta: do this ref count increment before creating the thread

src/replication.c
```diff
@@ -351,50 +375,40 @@ void clear_replica_cmd(spec_t *spec, replica_t *replica, rcmd_t *rep_cmd) {
...
-			close(replica->iofd);
-			close(replica->sender_epfd);
-		}
+	MTX_LOCK(&replica->r_mtx);
+	REPLICA_LOG("REMOVE REPLICA FROM LIST\n");
+	if(replica->state == REPLICA_DEGRADED)
+		spec->degraded_rcount--;
+	if(replica->state == REPLICA_HEALTHY)
+		spec->healthy_rcount--;
+	update_volstate(spec);
```
vishnuitta: do tailq_remove of replica from rq_list, and unlock the spec->rq_mtx here itself, but, can be before taking r_mtx lock

src/replication.c
```diff
@@ -351,50 +375,40 @@ void clear_replica_cmd(spec_t *spec, replica_t *replica, rcmd_t *rep_cmd) {
...
+		TAILQ_REMOVE(&replica->blockedq, rep_cmd, rblocked_cmd_next);
+		ios_aborted++;
 	}
+	pthread_cond_signal(&replica->r_cond);
+	TAILQ_REMOVE(&spec->rq, replica, r_next);
 	MTX_UNLOCK(&spec->rq_mtx);
-	REPLICA_LOG("%d IOs aborted for replica %d", ios_aborted, replica->id);
-	return 0;
+	rep_id = replica->id;
+	MTX_LOCK(&replica->r_mtx);
```
vishnuitta: taking r_mtx lock twice

src/replication.c
```diff
@@ -687,17 +730,61 @@ replica_receiver(void *arg) {
...
 	replica->mgmt_fd = mgmt_fd;
+	replica->iofd = -1;
+	replica->io_resp_hdr = NULL;
+	replica->ip = NULL;
 	replica->mgmt_io_state = READ_IO_RESP_HDR;
 	replica->mgmt_io_read = 0;
 	replica->mgmt_ack = (zvol_io_hdr_t *)malloc(sizeof(zvol_io_hdr_t));
 	replica->mgmt_ack_data = (mgmt_ack_data_t *)malloc(sizeof(mgmt_ack_data_t));
 	MTX_LOCK(&spec->rq_mtx);
+	replica->id = allocate_replica_id(spec);
```
vishnuitta: with latest changes, replica 'id' may not be required. While merging, this can be removed completely

vishnuitta: While merging, this can be removed completely if its not in use anymore

src/replication.c
```diff
@@ -713,15 +800,21 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd, char *replicaip
 
 	int rc;
 	pthread_t replica_sender_thread;
-	rc = pthread_mutex_init(&replica->r_mtx, NULL); //check
+	__sync_fetch_and_add(&replica->refcount, 1);
```
vishnuitta: why do we need this here? we already incremented in create_replica_entry right..?

src/replication.c
vishnuitta: basically, follow the reverse order of adding replica entry (i.e., on update_replica_entry).. so, it should be:
- remove from list, update vol state
- stop the threads
- close the sockets

src/replication.c
```diff
@@ -351,50 +375,40 @@ void clear_replica_cmd(spec_t *spec, replica_t *replica, rcmd_t *rep_cmd) {
...
+		ios_aborted++;
 	}
+	pthread_cond_signal(&replica->r_cond);
+	TAILQ_REMOVE(&spec->rq, replica, r_next);
 	MTX_UNLOCK(&spec->rq_mtx);
-	REPLICA_LOG("%d IOs aborted for replica %d", ios_aborted, replica->id);
-	return 0;
+	rep_id = replica->id;
+	MTX_LOCK(&replica->r_mtx);
+	__sync_fetch_and_sub(&replica->refcount, 1);
```
vishnuitta: cleanup_replica does the decrement.. is this for extra increment in add_replica_entry?
please have comments in explaining the corresponding increment for a decrement..

src/replication.c
vishnuitta: lets have protocol that a replica will be in spec's list only if it is in DEGRADED or HEALHTY state..
so, I mean that - replica need to be removed from list before changing its state to ERRORED

src/replication.c
```diff
@@ -625,59 +639,88 @@ typedef struct read_event {
...
-		}
 		//Wait for events on all iofds
 		event_count = epoll_wait(epfd, events, MAXEVENTS, -1);
 		for(i=0; i< event_count; i++) {
 			//Remove the replica from queue
 			if ((events[i].events & EPOLLERR) ||
 					(events[i].events & EPOLLHUP) ||
 					(!(events[i].events & EPOLLIN))) {
-				remove_replica_from_list(spec, events[i].data.fd);
+				get_replica_from_iofd(spec, events[i].data.fd);
```
vishnuitta: assign it to replica.. also, call remove_replica_from_list only if replica is not null

src/replication.c
vishnuitta: or is this decrement for removing from spec's rq list?
If I understood correctly, caller of this 'remove_replica_from_list' will also have a refcount incremented, and decrement related to that is done in cleanup_replica.. right?
some comments around this will be helpful

src/replication.c
```diff
@@ -625,59 +639,88 @@ typedef struct read_event {
...
+				remove_replica_from_list(spec, replica);
 				REPLICA_LOG("epoll error\n");
 				continue;
 			} else {
 				MTX_LOCK(&spec->rq_mtx);
 				TAILQ_FOREACH(replica, &spec->rq, r_next) {
 					if(events[i].data.fd == replica->iofd) {
+						MTX_LOCK(&replica->r_mtx);
+						if(replica->state == REPLICA_ERRORED) {
```
vishnuitta: this check is not required if we make sure that we remove from list before changing state to ERRORED..
so, this need to be removed based on changes in remove_replica_from_list

src/replication.c
```diff
@@ -625,59 +639,88 @@ typedef struct read_event {
...
+					MTX_UNLOCK(&replica->r_mtx);
+					cleanup_replica(replica);
+					continue;
+				} else if (ret == -1) {
+					MTX_UNLOCK(&replica->r_mtx);
+					remove_replica_from_list(spec, replica);
+					continue;
+				} else if (ret != 0) {
 					unblock_blocked_cmds(replica);
+					__sync_fetch_and_sub(&replica->refcount, 1);
```
vishnuitta: add a print that 'refcount' cant become 0 after decrement

src/replication.c
```diff
@@ -880,7 +975,9 @@ accept_mgmt_conns(int epfd, int sfd) {
 			close(mgmt_fd);
 			continue;
 		}
-		create_replica_entry(spec, mgmt_fd);
+		if((replica = create_replica_entry(spec, mgmt_fd)) == NULL) {
+			continue;
```
vishnuitta: we need to close mgmt_fd also

vishnuitta:
PR #19 addresses this issue of cleanup of replica. Hence, closing this one.

PR #12, opened 2018-04-17
Cstor#137 : maintain healthy and degraded replica count in target

mynktl:
- Added support for querying connected replica for status with opcode `ZVOL_OPCODE_REPLICA_STATUS` periodically
- According to replica status, ISTGT target will update healthy and degrade replica count.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -88,6 +88,23 @@ extern size_t rcommon_cmd_mempool_count;
 	MTX_UNLOCK(&spec->luworker_rmutex[luworker_id]); \
 }
 
+#define	clear_mgmt_cmd(_replica, _mgmt_cmd)		\
+	do {						\
+		TAILQ_REMOVE(&_replica->mgmt_cmd_queue,	\
+			_mgmt_cmd, mgmt_cmd_next);	\
+		free(_mgmt_cmd->io_hdr);		\
+		free(_mgmt_cmd->data);			\
+		free(_mgmt_cmd);			\
+	} while(0)
+
+#define	EMPTY_MGMT_Q_OF_REPLICA(_replica)					\
+	do {									\
+		mgmt_cmd_t *_mgmt_cmd;						\
+		while((_mgmt_cmd = TAILQ_FIRST(&replica->mgmt_cmd_queue))) {	\
+			clear_mgmt_cmd(_replica, _mgmt_cmd);			\
+		}								\
+        } while(0)
+
 int
 send_mgmtio(int fd, zvol_op_code_t mgmt_opcode, void *buf, size_t data_len) {
 	zvol_io_hdr_t *rmgmtio = NULL;
@@ -400,7 +417,7 @@ remove_replica_from_list(spec_t *spec, int iofd) {
 			EMPTY_Q_OF_REPLICA((&replica->waitq), rwait_cmd_next)
 			EMPTY_Q_OF_REPLICA((&replica->blockedq), rblocked_cmd_next)
 			EMPTY_Q_OF_REPLICA((&replica->read_waitq), rread_cmd_next)
-
+			EMPTY_MGMT_Q_OF_REPLICA(replica);
 			replica->removed = true;
 			pthread_cond_signal(&replica->r_cond);
 			MTX_UNLOCK(&replica->r_mtx);
@@ -464,6 +481,50 @@ read_data(int fd, uint8_t *data, uint64_t len, int *errorno, int *fd_closed) {
 		if(nbytes == len) {
 			break;
 		}
+	}
+	return nbytes;
+}
+
+/*
+ * write on fd for 'len'
+ * sets 'errorno' if write returns < 0
+ * closes fd if write returns < 0 && errno != EAGAIN, and sets fd_closed
+ * also closes fd if read return 0, i.e., EOF
+ * returns number of bytes written
+ */
+static int64_t
+write_data(int fd, uint8_t *data, uint64_t len, int *errorno, int *fd_closed) {
+	int64_t rc;
+	uint64_t nbytes = 0;
+
+	if (fd_closed != NULL)
+		*fd_closed = 0;
+	if (errorno != NULL)
+		*errorno = 0;
+
+	while(1) {
+		rc = write(fd, data + nbytes, len - nbytes);
+		if(rc < 0) {
+			if (errorno != NULL)
+				*errorno = errno;
+			//what about EINTR?
+			if(errno == EAGAIN || errno == EWOULDBLOCK)
+				return nbytes;
+			else {
+				REPLICA_ERRLOG("received err %d on fd %d, closing it..\n", errno, fd);
+				close(fd);
+				if (fd_closed != NULL)
+					*fd_closed = 1;
+				return -1;
+			}
+		} else if (rc == 0) {
+			REPLICA_ERRLOG("written 0 bytes on fd %d ..\n", fd);
+			break;
+		}
+
+		nbytes += rc;
+		if(nbytes == len)
+			break;
 	}
 	return nbytes;
 }
@@ -649,15 +710,15 @@ handle_write_resp(spec_t *spec, replica_t *replica)
 	return 0;
 }
 
-typedef struct read_event {
+typedef struct io_event {
 	int fd;
 	int *state;
-	zvol_io_hdr_t *resp_hdr;
-	void **resp_data;
-	int *read_count;
-} read_event_t;
+	zvol_io_hdr_t *io_hdr;
+	void **io_data;
+	int *byte_count;
+} io_event_t;
 
-int read_io_resp(spec_t *spec, replica_t *replica, read_event_t *revent);
+int read_io_resp(spec_t *spec, replica_t *replica, io_event_t *revent);
 
 //Receive IO responses in this thread
 void *
@@ -666,7 +727,7 @@ replica_receiver(void *arg) {
 	replica_t *replica;
 	int i, event_count, replica_count = 0, ret;
 	struct epoll_event event, *events;
-	read_event_t revent;
+	io_event_t revent;
 
 	//Create a new epoll epfd
 	int epfd = spec->receiver_epfd;
@@ -699,9 +760,9 @@ replica_receiver(void *arg) {
 
 				revent.fd = events[i].data.fd;
 				revent.state = &(replica->io_state);
-				revent.resp_hdr = replica->io_resp_hdr;
-				revent.resp_data = &(replica->io_resp_data);
-				revent.read_count = &(replica->io_read);
+				revent.io_hdr = replica->io_resp_hdr;
+				revent.io_data = &(replica->io_resp_data);
+				revent.byte_count = &(replica->io_read);
 
 				ret = read_io_resp(spec, replica, &revent);
 
@@ -721,28 +782,27 @@ replica_receiver(void *arg) {
 replica_t *
 create_replica_entry(spec_t *spec, int mgmt_fd)
 {
-	replica_t *replica = (replica_t *)malloc(sizeof(replica_t));
+	replica_t *replica = NULL;
+	int rc;
+
+	replica = (replica_t *)malloc(sizeof(replica_t));
+	if (!replica)
+		return NULL;
+
 	replica->mgmt_fd = mgmt_fd;
-	replica->mgmt_io_state = READ_IO_RESP_HDR;
-	replica->mgmt_io_read = 0;
-	replica->mgmt_ack = (zvol_io_hdr_t *)malloc(sizeof(zvol_io_hdr_t));
-	replica->mgmt_ack_data = (mgmt_ack_data_t *)malloc(sizeof(mgmt_ack_data_t));
+	TAILQ_INIT(&(replica->mgmt_cmd_queue));
 	MTX_LOCK(&spec->rq_mtx);
 	TAILQ_INSERT_TAIL(&spec->rwaitq, replica, r_waitnext);
 	MTX_UNLOCK(&spec->rq_mtx);
-	return replica;
-}
+	/*
+	 * As of now, we are supporting single spec_t per target
+	 * So, we can assign spec to replica here.
+	 * TODO: In case of multiple spec, asignment of spec to replica
+	 * 	 should be handled in update_replica_entry func according to
+	 * 	 volume name provided by replica.
+	 */
+	replica->spec = spec;
 
-/*
- * updates replica entry with IP/port
- * removes from spec's rwaitq, adds to rq after data connection
- * starts sender thread to send IOs to replica
- */
-replica_t *
-update_replica_entry(spec_t *spec, replica_t *replica, int iofd, char *replicaip, int replica_port) {
-
-	int rc;
-	pthread_t replica_sender_thread;
 	rc = pthread_mutex_init(&replica->r_mtx, NULL); //check
 	if (rc != 0) {
 		REPLICA_ERRLOG("pthread_mutex_init() failed errno:%d\n", errno);
@@ -753,10 +813,28 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd, char *replicaip
 		REPLICA_ERRLOG("pthread_cond_init() failed errno:%d\n", errno);
 		return NULL;
 	}
+	return replica;
+}
+
+/*
+ * updates replica entry with IP/port
+ * removes from spec's rwaitq, adds to rq after data connection
+ * starts sender thread to send IOs to replica
+ */
+replica_t *
+update_replica_entry(spec_t *spec, replica_t *replica, int iofd, char *replicaip, int replica_port) {
+	int rc;
+	zvol_io_hdr_t *rio;
+	pthread_t replica_sender_thread;
+	struct epoll_event event;
+
+	MTX_LOCK(&replica->r_mtx);
+
 	TAILQ_INIT(&replica->sendq);
 	TAILQ_INIT(&replica->waitq);
 	TAILQ_INIT(&replica->read_waitq);
 	TAILQ_INIT(&replica->blockedq);
+
 	replica->iofd = iofd;
 	replica->ip = malloc(strlen(replicaip)+1);
 	strcpy(replica->ip, replicaip);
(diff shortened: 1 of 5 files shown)


src/replication.c
```diff
@@ -814,6 +815,71 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd, char *replicaip
...
+	}
+
+	free(buf);
+}
+
+static int
+update_replica_status(spec_t *spec, replica_t *replica)
+{
+	replica_state_t last_status;
+	zrepl_status_ack_t *repl_status = (zrepl_status_ack_t *) replica->mgmt_ack_data;
```
vishnuitta: need to free memory for mgmt_ack_data

src/replication.c
```diff
@@ -899,15 +1095,23 @@ accept_mgmt_conns(int epfd, int sfd) {
...
+			close(mgmt_fd);
+			continue;
+		}
+		//Hack to handle read of ZVOL_OPCODE_HANDSHAKE cmd
+		queue_handshake_read_msg(replica);
+		event.data.ptr = replica;
+		event.events = EPOLLIN | EPOLLHUP | EPOLLERR | EPOLLET | EPOLLOUT;
 		rc = epoll_ctl(epfd, EPOLL_CTL_ADD, mgmt_fd, &event);
 		if(rc == -1) {
 			REPLICA_ERRLOG("epoll_ctl() failed on fd %d, errno:%d.. closing it..", mgmt_fd, errno);
```
vishnuitta: cleanup of replica need to be done

src/replication.c
```diff
@@ -1030,28 +1267,81 @@ read_io_resp(spec_t *spec, replica_t *replica, read_event_t *revent)
...
 
-	replica_t *replica = get_replica(fd, &spec);
-	if (replica == NULL || spec == NULL) {
-		REPLICA_ERRLOG("error in getting replica %p/spec %p for fd %d..\n", replica, spec, fd);
+	MTX_LOCK(&replica->r_mtx);
+	mgmt_cmd = TAILQ_FIRST(&replica->mgmt_cmd_queue);
+	if (!(mgmt_cmd != NULL &&
+		(mgmt_cmd->mgmt_cmd_state == READ_IO_RESP_HDR ||
+		mgmt_cmd->mgmt_cmd_state == READ_IO_RESP_DATA))) {
+		MTX_UNLOCK(&replica->r_mtx);
```
vishnuitta: this need to be handled, and a print is required.. probably, by doing 'exit'

src/replication.c
```diff
@@ -986,14 +1205,14 @@ read_io_resp(spec_t *spec, replica_t *replica, read_event_t *revent)
 
 			*read_count = 0;
 			if (resp_hdr->len != 0) {
-			/* this will not be NULL for mgmt_ack IO */
 				if ((*resp_data) == NULL)
 					(*resp_data) = malloc(resp_hdr->len);
 			}
-			*state = READ_IO_RESP_DATA;
```
vishnuitta: this place is correct right for setting the state..

src/replication.c
```diff
@@ -720,15 +782,37 @@ replica_receiver(void *arg) {
...
-	replica->mgmt_io_state = READ_IO_RESP_HDR;
-	replica->mgmt_io_read = 0;
-	replica->mgmt_ack = (zvol_io_hdr_t *)malloc(sizeof(zvol_io_hdr_t));
-	replica->mgmt_ack_data = (mgmt_ack_data_t *)malloc(sizeof(mgmt_ack_data_t));
+	TAILQ_INIT(&(replica->mgmt_cmd_queue));
 	MTX_LOCK(&spec->rq_mtx);
 	TAILQ_INSERT_TAIL(&spec->rwaitq, replica, r_waitnext);
 	MTX_UNLOCK(&spec->rq_mtx);
+	/*
+	 * As of now, we are supporting single spec_t per target
```
vishnuitta: this comment is not required here.. its required in accept_mgmt_conns which passes the spec..

src/replication.c
```diff
@@ -739,23 +823,18 @@ create_replica_entry(spec_t *spec, int mgmt_fd)
...
-		return NULL;
-	}
-	rc = pthread_cond_init(&replica->r_cond, NULL); //check
-	if (rc != 0) {
-		REPLICA_ERRLOG("pthread_cond_init() failed errno:%d\n", errno);
-		return NULL;
-	}
+	struct epoll_event event;
+
+	MTX_LOCK(&replica->r_mtx);
```
vishnuitta: replica's r_mtx can be taken only after spec's rq_mtx.
And, why do we need lock here?

src/replication.c
```diff
@@ -1030,28 +1267,81 @@ read_io_resp(spec_t *spec, replica_t *replica, read_event_t *revent)
...
-	revent.fd = fd;
-	revent.state = &(replica->mgmt_io_state);
-	revent.resp_hdr = replica->mgmt_ack;
-	revent.resp_data = (void **)(&(replica->mgmt_ack_data));
-	revent.read_count = &(replica->mgmt_io_read);
+	MTX_UNLOCK(&replica->r_mtx);
+	int count = 0;
+
+	revent.fd = replica->mgmt_fd;
+	revent.state = &(mgmt_cmd->mgmt_cmd_state);
```
vishnuitta: we can probably use mgmt_cmd_s itself rather than copying it to revent structure..

src/replication.c
```diff
@@ -1030,28 +1267,81 @@ read_io_resp(spec_t *spec, replica_t *replica, read_event_t *revent)
...
+	revent.io_hdr = mgmt_cmd->io_hdr;
+	revent.io_data = (void **)(&(mgmt_cmd->data));
+	revent.byte_count = &(mgmt_cmd->io_bytes);
 
-	read_io_resp(spec, replica, &revent);
+	count = read_io_resp(replica->spec, replica, &revent);
+	if (count) {
+		clear_mgmt_cmd(replica, mgmt_cmd);
```
vishnuitta: would be needing r_mtx lock here

src/replication.c
```diff
@@ -1012,15 +1232,32 @@ read_io_resp(spec_t *spec, replica_t *replica, read_event_t *revent)
...
+					if (fd != replica->iofd)
+						update_replica_status(spec, replica, *resp_data);
+					free(*resp_data);
+					*resp_data = NULL;
+					break;
+				default:
+					REPLICA_NOTICELOG("unsupported opcode(%d) received..\n", resp_hdr->opcode);
 					break;
 			}
+
```
vishnuitta: should we set *resp_data to NULL here which is common to all places..

src/replication.c
```diff
@@ -926,53 +1130,68 @@ accept_mgmt_conns(int epfd, int sfd) {
...
+	int donecount = 0;
+
+	switch(*state) {
+		case WRITE_IO_SEND_HDR:
+			reqlen = sizeof (zvol_io_hdr_t) - (*write_count);
+			count = write_data(fd, ((uint8_t *)write_hdr) + (*write_count), reqlen, &errorno, &fd_closed);
+			CHECK_AND_ADD_BREAK_IF_PARTIAL(fd_closed, (*write_count), count, reqlen);
+
+			*write_count = 0;
+			*state = WRITE_IO_SEND_DATA;
```
vishnuitta: we should go to WRITE_IO_SEND_DATA case if there is nothing to write

src/replication.c
```diff
@@ -926,53 +1130,68 @@ accept_mgmt_conns(int epfd, int sfd) {
...
 			}
-		}
-		if (replica != NULL) {
-			MTX_UNLOCK(&spec->rq_mtx);
-			break;
-		}
-		TAILQ_FOREACH(replica, &spec->rwaitq, r_waitnext) {
-			if(replica->mgmt_fd == mgmt_fd) {
-				*s = spec;
+			if (errorno == EAGAIN || errorno == EWOULDBLOCK)
```
vishnuitta: this EAGAIN can be removed in case of WRITE. We can move to READ state.
And, if EPOLLOUT comes, we need not do anything as the IO in the head of list is waiting on READ state.

src/replication.c
```diff
@@ -1030,28 +1267,81 @@ read_io_resp(spec_t *spec, replica_t *replica, read_event_t *revent)
...
+	int rc = -1;
+	mgmt_cmd_t *mgmt_cmd;
+
+	MTX_LOCK(&replica->r_mtx);
+	mgmt_cmd = TAILQ_FIRST(&replica->mgmt_cmd_queue);
+	if (!(mgmt_cmd != NULL &&
+		(mgmt_cmd->mgmt_cmd_state == WRITE_IO_SEND_HDR ||
+		mgmt_cmd->mgmt_cmd_state == WRITE_IO_SEND_DATA))) {
+		MTX_UNLOCK(&replica->r_mtx);
+		return rc;
```
vishnuitta: lets add a comment that we hit this case when this IO in the head of list is waiting for read and in READ related state.

src/replication.c
```diff
@@ -788,48 +869,136 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd, char *replicaip
...
+
+	MTX_LOCK(&replica->r_mtx);
+	last_status = replica->state;
+	replica->state = repl_status->state;
+	MTX_UNLOCK(&replica->r_mtx);
+
+	if(last_status != repl_status->state) {
+		MTX_LOCK(&spec->rq_mtx);
+		if ((replica->state == REPLICA_DEGRADED ||
+			replica->state == REPLICA_ERRORED) && last_status == REPLICA_HELATHY) {
```
vishnuitta: don't use replica->state here, use repl_status->state.
we don't need to use ERRORED. ERRORED state should actually get disconnected.

src/replication.c
```diff
@@ -839,11 +1008,36 @@ zvol_handshake(spec_t *spec, replica_t *replica)
...
 	return 0;
 }
 
+/*
+ * queue_handshake_read_msg add mgmt_cmd related to HANDSHAKE
+ * (which is already sent to replica) in mgmt_cmd_queue of replica
+ * for upcoming read
+ */
+static void
+queue_handshake_read_msg(replica_t *replica)
```
vishnuitta: why don't we queue for sending handshake msg?

src/replication.c
```diff
@@ -1030,28 +1267,81 @@ read_io_resp(spec_t *spec, replica_t *replica, read_event_t *revent)
...
+
+	revent.fd = replica->mgmt_fd;
+	revent.state = &(mgmt_cmd->mgmt_cmd_state);
+	revent.io_hdr = mgmt_cmd->io_hdr;
+	revent.io_data = (void **)(&(mgmt_cmd->data));
+	revent.byte_count = &(mgmt_cmd->io_bytes);
 
-	read_io_resp(spec, replica, &revent);
+	count = read_io_resp(replica->spec, replica, &revent);
+	if (count) {
```
vishnuitta: lets have a print if it is more than 1..

src/replication.c
```diff
@@ -1030,28 +1267,81 @@ read_io_resp(spec_t *spec, replica_t *replica, read_event_t *revent)
...
+		clear_mgmt_cmd(replica, mgmt_cmd);
+	}
```
vishnuitta: we need to take next IO in the queue to send the data

src/replication.c
```diff
@@ -467,6 +485,50 @@ read_data(int fd, uint8_t *data, uint64_t len, int *errorno, int *fd_closed) {
...
+		*fd_closed = 0;
+	if (errorno != NULL)
+		*errorno = 0;
+
+	while(1) {
+		rc = write(fd, data + nbytes, len - nbytes);
+		if(rc < 0) {
+			if (errorno != NULL)
+				*errorno = errno;
+			//what about EINTR?
```
vishnuitta: yes EINTR need to be handled.. for read case also..

src/replication.c
```diff
@@ -899,15 +1095,23 @@ accept_mgmt_conns(int epfd, int sfd) {
...
+		replica = create_replica_entry(spec, mgmt_fd);
+		if (!replica) {
+			REPLICA_ERRLOG("Failed to create replica for fd %dclosing it..", mgmt_fd);
+			close(mgmt_fd);
+			continue;
+		}
+		//Hack to handle read of ZVOL_OPCODE_HANDSHAKE cmd
+		queue_handshake_read_msg(replica);
+		event.data.ptr = replica;
+		event.events = EPOLLIN | EPOLLHUP | EPOLLERR | EPOLLET | EPOLLOUT;
```
vishnuitta: EPOLLRDHUP need to be added here and in the init_replication

src/replication.c
mynktl: there was a separate function send_mgmtio for sending handshake message so I have avoided queueing handshake msg. We can surely queue handshake message in mgmt_cmd_queue. I will update it accordingly.

src/replication.c
mynktl: We are not using a break here, so it will go to next case if CHECK_AND_ADD_BREAK_IF_PARTIAL passes.

src/replication.c
mynktl: ok. I was thinking do we need to check for errorno after read_data as read_data itself will give count lesser than reqlen if read_data hit any error during read operation.

src/replication.c

src/replication.c
```diff
@@ -463,95 +421,70 @@ remove_replica_from_list(spec_t *spec, int iofd) {
...
+				break;
+		}
+
 		if(rc < 0) {
 			if (errorno != NULL)
 				*errorno = errno;
-			if(errno == EAGAIN || errno == EWOULDBLOCK)
+
+			if (errno == EAGAIN || errno == EWOULDBLOCK ||
+			    errno == EINTR) {
```
vishnuitta: for EINTR, we need to do contine

src/replication.c
```diff
@@ -463,95 +421,70 @@ remove_replica_from_list(spec_t *spec, int iofd) {
...
 				return nbytes;
-			else {
+			} else {
 				REPLICA_ERRLOG("received err %d on fd %d, closing it..\n", errno, fd);
-				/*
```
vishnuitta: lets leave the comment as it is

src/replication.c
```diff
@@ -463,95 +421,70 @@ remove_replica_from_list(spec_t *spec, int iofd) {
...
-				 * TODO: cleanup of replica need to be happen
-				 */
 				close(fd);
 				if (fd_closed != NULL)
 					*fd_closed = 1;
 				return -1;
 			}
-		}
-		else if (rc == 0) {
+		} else if (rc == 0) {
```
vishnuitta: 'write' call returning 0 doesn't mean error.. so, that need to be handled differently..

src/replication.c
```diff
@@ -936,16 +873,13 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd, char *replicaip
 	write(replica->iofd, spec->lu->volname, strlen(spec->lu->volname));
 	free(rio);
 
-	update_volstate(spec);
-
 	event.data.fd = replica->iofd;
 	event.events = EPOLLIN | EPOLLET;
 	rc = epoll_ctl(spec->receiver_epfd, EPOLL_CTL_ADD, iofd, &event);
 	if (rc == -1) {
 		return NULL;
```
vishnuitta: need to do a cleanup here of updating volstate again. Lets a add a 'TODO' here, and also a bug.
Similarly, for below cases also.

src/replication.c
```diff
@@ -857,53 +872,215 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
...
 	}
+
 	replica->sender_epfd = epoll_create1(0);
 	event.data.fd = replica->iofd;
 	event.events = EPOLLOUT;
 	rc = epoll_ctl(replica->sender_epfd, EPOLL_CTL_ADD, replica->iofd, &event);
 	if (rc == -1) {
 		REPLICA_LOG("epoll_ctl_add failed errno:%d\n", errno);
+		MTX_UNLOCK(&replica->r_mtx);
```
vishnuitta: this need to be removed

src/replication.c
```diff
@@ -967,6 +900,77 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd, char *replicaip
 	return replica;
 }
 
+static void
```
vishnuitta: did you picked changes from other pending PR from @payes ?

src/replication.c
```diff
@@ -967,6 +900,77 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd, char *replicaip
...
+	mgmt_cmd->io_hdr = rmgmtio;
+	mgmt_cmd->io_bytes = 0;
+	mgmt_cmd->data = data;
+	mgmt_cmd->mgmt_cmd_state = WRITE_IO_SEND_HDR;
+
+	MTX_LOCK(&replica->r_mtx);
+	TAILQ_INSERT_TAIL(&replica->mgmt_cmd_queue, mgmt_cmd, mgmt_cmd_next);
+	MTX_UNLOCK(&replica->r_mtx);
+
+	handle_write_data_event(replica);
```
vishnuitta: we don't need this. we get a EPOLLOUT in the starting.

src/replication.c

src/replication.c
mynktl: yes.

src/replication.c
mynktl: Yes. To avoid conflicts, I pulled Payes changes for cleanup_replica so we can have a single version of cleanup_replica.

src/replication.c

src/replication.c
```diff
@@ -851,16 +784,12 @@ create_replica_entry(spec_t *spec, int mgmt_fd)
 
 	replica->mgmt_fd = mgmt_fd;
 	TAILQ_INIT(&(replica->mgmt_cmd_queue));
+	replica->initial_checkpointed_io_seq = 0;
+	replica->mgmt_io_resp_hdr = malloc(sizeof(zvol_io_hdr_t));
```
vishnuitta: mgmt_io_resp_data should be set to NULL

src/replication.c
vishnuitta: sorry.. this comment is at wrong place.. we need to call this fn here..
I wanted to comment at the place where handshake command was made

src/replication.c
```diff
@@ -857,53 +882,222 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
...
+	repl_status = (zrepl_status_ack_t *)replica->mgmt_io_resp_data;
+
+	MTX_LOCK(&replica->r_mtx);
+	last_status = replica->state;
+	replica->state = (replica_state_t) repl_status->state;
+	MTX_UNLOCK(&replica->r_mtx);
+
+	if(last_status != repl_status->state) {
+		MTX_LOCK(&spec->rq_mtx);
+		if (replica->state == ZVOL_STATUS_DEGRADED) {
```
vishnuitta: lets use repl_status state rather than replica's state

PR #14, opened 2018-05-04
Handling of read ios to multiple replicas at target

payes:
In istgt, when healthy replicas exists, send read IOs only to healthy replicas.
If there are no healthy replicas, read IOs need to be sent for all degraded replicas. Once these read responses are received, response with highest IO number need to be served to client.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -157,17 +157,16 @@ dequeue_common_sendq:
 		MTX_UNLOCK(&spec->rcommonq_mtx);
 
 		//Enqueue to individual replica cmd queues and send on the respective fds
-		send_multiple_read = false;
 		if(cmd->opcode == ZVOL_OPCODE_READ)
 			cmd_read = true;
 		else
 			cmd_read = false;
 		MTX_LOCK(&spec->rq_mtx);
 		if(spec->healthy_rcount == 0) {
-			send_multiple_read = true;
+			send_multiple_read = cmd->send_multiple_read = true;
 			healthy_untried = 0;
 		} else {
-			send_multiple_read = false;
+			send_multiple_read = cmd->send_multiple_read = false;
 			healthy_untried = spec->healthy_rcount;
 		}
 
@@ -585,6 +584,19 @@ get_read_resp_data(rcommon_cmd_t *rcommq_ptr, uint64_t blocklen)
 	return read_data;
 }
 
+#define handle_read_healthy() { \
+	dataptr = get_read_resp_data(rcommq_ptr, spec->blocklen); \
+	TAILQ_REMOVE(&spec->rcommon_waitq, rcommq_ptr, wait_cmd_next); \
+	rcommq_ptr->state = CMD_EXECUTION_DONE; \
+	while(data_read = TAILQ_FIRST(&rcommq_ptr->data_read_ptr)) { \
+		free(data_read->io_resp_data_ptr_cpy); \
+		TAILQ_REMOVE(&rcommq_ptr->data_read_ptr, data_read, data_next); \
+		free(data_read); \
+	} \
+	rcommq_ptr->data = dataptr; \
+	signal_luworker(); \
+}
+
 #define handle_read_consistency_met() { \
 	dataptr = get_read_resp_data(rcommq_ptr, spec->blocklen); \
 	TAILQ_REMOVE(&spec->rcommon_waitq, rcommq_ptr, wait_cmd_next); \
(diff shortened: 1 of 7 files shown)


src/replication.c
```diff
@@ -157,15 +157,28 @@ replicator(void *arg) {
...
+			healthy_untried = spec->healthy_rcount;
+		}
 
 		TAILQ_FOREACH(replica, &spec->rq, r_next) {
 			if(replica == NULL) {
 				REPLICA_LOG("Replica not present");
 				MTX_UNLOCK(&spec->rq_mtx);
 				exit(EXIT_FAILURE);
 			}
+			if(cmd_read && !send_multiple_read && replica->state != ZVOL_STATUS_HEALTHY) {
```
vishnuitta: replica lock required here..

vishnuitta: it should be ok.. not required a lock here as long as replica data structure is intact

src/replication.c
```diff
@@ -185,16 +199,20 @@ replicator(void *arg) {
...
-					if(rcmd->opcode == ZVOL_OPCODE_READ) {
-						read_cmd_sent = true;
-					}
 					pthread_cond_signal(&replica->r_cond);
+					if(cmd_read && !send_multiple_read) {
+						MTX_UNLOCK(&replica->r_mtx);
+						break;
+					}
+				} else {
+					if(cmd_read && (--healthy_untried > 0)) {
```
vishnuitta: healthy_untried is set to healthy_count, but, its getting decremented if cmd is blocked, which doesn't seem to be correct.
You can just have logic where:
- find whether multiple sends are required or not by looking at healthy_rcount of spec
- if multiple need to be sent, just keep sending to all by adding either to blockedq or sendq
- if multiple need not to be sent, just keep searching for a healthy one, add to its blockedq or sendq and do break
- if no such healthy one found, just iterate once again by setting multiple sends as required

src/replication.c
```diff
@@ -497,11 +520,13 @@ perform_read_write_on_fd(int fd, uint8_t *data, uint64_t len, int *errorno,
 	return nbytes;
 }
 
-void unblock_blocked_cmds(replica_t *replica)
+void
+unblock_blocked_cmds(replica_t *replica)
```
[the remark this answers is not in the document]
vishnuitta: ok.. this is also fine.. this is another way of finding any blockages due to ongoing IOs. We just need to look at rwaitq also as mentioned in the comment of this function.

src/replication.c
```diff
@@ -567,47 +651,50 @@ handle_read_resp(spec_t *spec, replica_t *replica)
...
+			luworker_id = rcommq_ptr->luworker_id;
+			if(!rcommq_ptr->send_multiple_read) {
+				handle_read_healthy();
+			} else if (rcommq_ptr->acks_recvd == MAX(spec->replica_count - spec->consistency_factor + 1,
+								spec->consistency_factor)) {
+				handle_read_consistency_met();
 				MTX_UNLOCK(&spec->rcommonq_mtx);
-				clear_rcomm_cmd(rcommq_ptr);
+				MTX_UNLOCK(&rcommq_ptr->rcommand_mtx);
+			} else if (rcommq_ptr->acks_recvd + rcommq_ptr->ios_aborted
```
vishnuitta: we can just have the case for 'acks_recvd + ios_aborted == copies_sent'. In that also, we should be able to handle by going through all the responses received.
Reason for this approach is: There might be change in the state of replica after sending the requests to replicas, like, after sending the request, replica's rebuilding might have got completed and may not be sending metadata. So, after receiving all the responses, do:
- check for any response with io_num as 0, and send that as the response
- else, check whether acks_recvd is equal to MAX(.....), and build the response and send that
- else, send error response
Finally, do the cleanup of the command and its memory.

src/replication.c
```diff
@@ -527,38 +552,97 @@ void unblock_blocked_cmds(replica_t *replica)
...
-		dataptr += (io_hdr->len);
-		parsed += (sizeof(struct zvol_io_rw_hdr) + io_hdr->len);
+	int64_t io_num;
+	int64_t nbytes = 0;
+	uint8_t *tmp_data = NULL;
+	data_read_t *dataptr = NULL;
+	struct zvol_io_rw_hdr *io_hdr = NULL;
+	uint8_t *read_data = (uint8_t *)malloc(rcommq_ptr->data_len);
+	while(nbytes < rcommq_ptr->data_len) {
+		io_num = -1;
```
vishnuitta: better not to use -ve numbers, as io_num from replica is a uint64

src/replication.c
```diff
@@ -527,38 +552,97 @@ void unblock_blocked_cmds(replica_t *replica)
...
-		memcpy(read_data + len, dataptr, io_hdr->len);
-		len += io_hdr->len;
-		dataptr += (io_hdr->len);
-		parsed += (sizeof(struct zvol_io_rw_hdr) + io_hdr->len);
+	int64_t io_num;
+	int64_t nbytes = 0;
+	uint8_t *tmp_data = NULL;
+	data_read_t *dataptr = NULL;
+	struct zvol_io_rw_hdr *io_hdr = NULL;
+	uint8_t *read_data = (uint8_t *)malloc(rcommq_ptr->data_len);
```
vishnuitta: should we do xmalloci..?

src/replication.c
payes: If a response has io_num as 0 in the IO header, how do we find that IO in our queues. I went through the zrepl code, it looks like we need to add one more variable to our IO header to use this logic or use the flag variable.

vishnuitta:
This is handled through different PR https://github.com/openebs/istgt/pull/35.
Hence, closing this PR.

PR #17, opened 2018-05-15
Fix stack smash error in istgt

utkarshmani1997:
On branch fix-stack-smashing
Changes to be committed:
   modified:   src/replication_test.c

This commit fixes the stack smash error in istgt

Signed-off-by: Utkarsh Mani Tripathi <utkarshmani1997@gmail.com>


vishnuitta:
thanks Utkarsh

PR #18, opened 2018-05-16
[US1537] snapshot related tasks

vishnuitta:
This PR addresses the below snapshot related tasks:
- implement 'snapcreate' and 'snapdestroy' options for istgtcontrol
- API to pause write IOs, waits for any ongoing write IOs to finish
- API to send snapshot create/destroy commands to replica
- API to create and destroy snapshots by using all above functions

Testcases for this feature
- Verifying if volume exists for correct volname
- Verifying if volume does not exist for wrong volname
- Verifying status of volume as degraded with less number of replicas connected
- Verifying volume state is healthy with required number of replicas connected and being in healthy state
- Hitting the case of timeout while waiting for ongoing write IOs being served, and verify that write IOs are resumed and error is returned from API
- Verifying that configured amount of time has not been passed by the time snapshot IO is being sent
- Hit the case of adding snapshot command to the list, and verify command is sent eventually
- Verify that snapshots are destroyed in case of one or multiple failures from replica
- Verify that write IOs are resumed in case of one or multiple failures from replica
- Hit the case of timeout while waiting for snapshot response(s) and verify that write IOs are resumed, volumes are destroyed and error is returned
- Verify success case of creating snapshot, write IOs are resumed and returned 0 as success from API

Created "TS2" in rally to track more testcases

This PR also addresses:
- added to blocked q incase of any blockages
- lock issue while removing replica from list
- updating healthy/degraded count while removing replica from list
- breaking out of loop if write returns 0 while performing write on a fd
- handling error response while performing read/write on fd
- clean up of replica during error in handshake response

This PR is not handling the case of erroring out snapshot command in case of replica close.
Updated this detail in US1536


vishnuitta:
This PR is outdated one. Tracking this feature with other PR #26.
Hence, closing this one.

PR #19, opened 2018-05-24
DE18 : marking replica as down in target

mynktl:
This is an initial version of this PR which addresses 
         1. Error handling in ISTGT's replica interface (if the replica get disconnected)
          2. Optimization in LUN worker and replica interface regarding the handling of command queues.

### DIFF
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -120,7 +120,7 @@ inform_mgmt_conn(replica_t *r)
 {
 	uint64_t num = 1;
 	r->disconnect_conn = 1;
-	write(r->mgmt_eventfd1, &num, sizeof (num));
+	(void) write(r->mgmt_eventfd1, &num, sizeof (num));
 }
 
 static rcmd_t *
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -120,7 +120,8 @@ inform_mgmt_conn(replica_t *r)
 {
 	uint64_t num = 1;
 	r->disconnect_conn = 1;
-	(void) write(r->mgmt_eventfd1, &num, sizeof (num));
+	if (write(r->mgmt_eventfd1, &num, sizeof (num)) != sizeof (num))
+		REPLICA_NOTICELOG("Failed report err to mgmt_conn for replica(%p)\n", r);
 }
 
 static rcmd_t *
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -1,0 +1,654 @@
+#include <sys/epoll.h>
+#include <sys/eventfd.h>
+#include <sys/prctl.h>
+#include <stdbool.h>
+#include "istgt.h"
+#include "zrepl_prot.h"
+#include "istgt_integration.h"
+#include "replication.h"
+#include "ring_mempool.h"
+
+#define	READ_PARTIAL	1
+#define	READ_COMPLETED	2
+
+#define	WRITE_PARTIAL	1
+#define	WRITE_COMPLETED	2
+
+extern rte_smempool_t rcmd_mempool;
+
+bool unblock_cmds(replica_t *r);
+void move_to_blocked_or_ready_q(replica_t *r, rcmd_t *cmd);
+void respond_with_error_for_all_outstanding_ios(replica_t *r);
+int handle_data_conn_error(replica_t *r);
+int handle_epoll_out_event(replica_t *r);
+static ssize_t perform_read_on_fd(int fd, uint8_t *data, uint64_t len);
+rcmd_t * find_replica_cmd(replica_t *r, uint64_t ioseq);
+int read_cmd(replica_t *r);
+int write_cmd(int fd, rcmd_t *cmd);
+int handle_data_eventfd(void *arg);
+int do_drainfd(int data_eventfd);
+int handle_epoll_in_event(replica_t *r);
+void *replica_thread(void *arg);
+
+#if 0
+	parsed = 0;
+	data_node = TAILQ_FIRST(&cmd->data_list);
+	while (data_node != NULL) {
+		next_node = TAILQ_NEXT(data_node);
+		ret = write_data_segment(data_node, parsed, written);
+		if (ret == WRITE_COMPLETED) {
+			data_node = next_node;
+			continue;
+		}
+		else
+			break;
+	}
+}
+
+write_status_t
+write_data_segment(data_node, parsed, written)
+{
+	len = data_node->len;
+	if ((parsed + len) <= written) {
+		parsed += len;
+		data_node = next_node;
+		continue;
+	}
+	offset_to_write = (written - parsed);
+	len_to_write = (parsed + len - written);
+	rc = write(data_node->data_ptr, offset_to_write, len_to_write);
+}
+#endif
+
+#define check_for_blockage() { \
+	uint64_t current_lba = cmd->offset; \
+	uint64_t current_lbE = current_lba + cmd->data_len; \
+	uint64_t pending_lba = pending_rcmd->offset; \
+	uint64_t pending_lbE = pending_lba + pending_rcmd->data_len; \
+	if ((current_lbE < pending_lba)  || (pending_lbE < current_lba)) { \
+		cmd_blocked = false; \
+	} else { \
+		cmd_blocked = true; \
+	} \
+}
+
+#define CHECK_BLOCKAGE_IN_Q(queue, next) { \
+	if(cmd_blocked == false) { \
+		TAILQ_FOREACH(pending_rcmd, queue, next) { \
+			check_for_blockage(); \
+			if(cmd_blocked == true) \
+				break; \
+		} \
+	} \
+}
+
+#define SEND_ERROR_RESPONSES(head) { \
+	cmd = TAILQ_FIRST(head); \
+	while (cmd != NULL) { \
+		next_cmd = TAILQ_NEXT(cmd, next); \
+		TAILQ_REMOVE(head, cmd, next); \
+		idx = cmd->idx; \
+		rcomm_cmd = cmd->rcommq_ptr; \
+		rcomm_cmd->resp_list[idx].io_resp_hdr.status = ZVOL_OP_STATUS_FAILED; \
+		rcomm_cmd->resp_list[idx].data_ptr = NULL; \
+		rcomm_cmd->resp_list[idx].data_len = 0; \
+		mtx = rcomm_cmd->mutex;			\
+		cond_var = rcomm_cmd->cond_var;		\
+		rcomm_cmd->resp_list[idx].status = -1; \
+		MTX_LOCK(mtx); \
+		pthread_cond_signal(cond_var); \
+		MTX_UNLOCK(mtx); \
+		cmd = next_cmd; \
+	} \
+}
+
+bool
+unblock_cmds(replica_t *r)
+{
+	bool cmd_blocked = false;
+	rcmd_t *pending_rcmd, *cmd, *next_cmd;
+	bool unblocked = false;
+
+	for (cmd = TAILQ_FIRST(&r->blockedq); cmd; cmd = next_cmd) {
+		next_cmd = TAILQ_NEXT(cmd, next);
+		cmd_blocked = false;
+		CHECK_BLOCKAGE_IN_Q(&r->readyq, next);
+		if (cmd_blocked == true)
+			continue;
+		CHECK_BLOCKAGE_IN_Q(&r->waitq, next);
+		if (cmd_blocked == true)
+			continue;
+		TAILQ_REMOVE(&r->blockedq, cmd, next);
+		TAILQ_INSERT_TAIL(&r->readyq, cmd, next);
+		unblocked = true;
+	}
+	return unblocked;
+}
+
+void
+move_to_blocked_or_ready_q(replica_t *r, rcmd_t *cmd)
+{
+	bool cmd_blocked = false;
+	rcmd_t *pending_rcmd;
+
+	if (!TAILQ_EMPTY(&r->blockedq)) {
+		TAILQ_INSERT_TAIL(&r->blockedq, cmd, next);
+		goto done;
+	}
+	CHECK_BLOCKAGE_IN_Q(&r->readyq, next);
+	CHECK_BLOCKAGE_IN_Q(&r->waitq, next);
+	if (cmd_blocked == true)
+		TAILQ_INSERT_TAIL(&r->blockedq, cmd, next);
+	else
+		TAILQ_INSERT_TAIL(&r->readyq, cmd, next);
+done:
+	return;
+}
+
+#if 0
+void
+inform_mgmt_conn(replica_t *r)
+{
+	uint64_t num = 10;
+
+	zvol_io_hdr_t *cmd;
+	cmd->opcode = DESTROY;
+	add_to_linked_list(&r->mgmt_cmd_list, cmd);
+	write(r->mgmt_event_fd, &num, sizeof (num));
+}
+#endif
+
+void *
+dequeue_replica_cmdq(replica_t *replica)
+{
+	unsigned count = 0;
+
+	count = get_num_entries_from_mempool(&(replica->cmdq));
+	if (count)
+		return get_from_mempool(&(replica->cmdq));
+	else
+		return NULL;
+}
+
+void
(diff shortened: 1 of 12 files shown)


vishnuitta:
Please fix the warnings also. There may be things which are not related to this PR. Focus on those that are related to modified files.

src/istgt_lu_disk.c
```diff
@@ -5552,68 +5564,214 @@ istgt_lu_disk_scsi_reserve(ISTGT_LU_DISK *spec, CONN_Ptr conn, ISTGT_LU_CMD_Ptr
...
-			MTX_UNLOCK(&rcomm_cmd->rcommand_mtx);
-			clear_rcomm_cmd(rcomm_cmd);
-		} else {
-			rcomm_cmd->completed = true;
-			MTX_UNLOCK(&rcomm_cmd->rcommand_mtx);
+
+	build_rcomm_cmd;
+
+	TAILQ_FOREACH(replica, &spec->rq, r_next) {
+		if(replica == NULL) {
```
vishnuitta: we don't need this

src/istgt_lu_disk.c
```diff
@@ -5552,68 +5564,214 @@ istgt_lu_disk_scsi_reserve(ISTGT_LU_DISK *spec, CONN_Ptr conn, ISTGT_LU_CMD_Ptr
...
+	for (i = 0; i < rcomm_cmd->copies_sent; i++) {
+		if (rcomm_cmd->resp_list[i].status == 1) {
+			success++;
+		} else if (rcomm_cmd->resp_list[i].status == -1) {
+			failure++;
+		}
+	}
+
+	if (rcomm_cmd->opcode == ZVOL_OPCODE_READ) {
+		if ((rcomm_cmd->copies_sent != (success + failure))) {
```
vishnuitta: we don't need to wait till all responses comes. If 'success' is reached to required number (i.e, either one success response from healthy replica or (n-m+1) success responses from degraded.. here, n is replication factor, m is consistency factor), response can be sent back to client.

src/istgt_lu_disk.c
```diff
@@ -5552,68 +5564,214 @@ istgt_lu_disk_scsi_reserve(ISTGT_LU_DISK *spec, CONN_Ptr conn, ISTGT_LU_CMD_Ptr
...
+static uint8_t *
+handle_read_consistency(rcommon_cmd_t *rcomm_cmd)
+{
+	int i;
+	uint8_t *dataptr = NULL;
+	struct io_data_chunk_list_t io_data_chunk_list;
+
+	TAILQ_INIT(&(io_data_chunk_list));
+
+	for (i = 0; i < rcomm_cmd->copies_sent; i++) {
```
vishnuitta: we should send success if success count reaches to required number as mentioned in comment at caller of this function.

src/replication.c
```diff
@@ -497,292 +180,72 @@ perform_read_write_on_fd(int fd, uint8_t *data, uint64_t len, int *errorno,
...
-	} else { \
-		rcommq_ptr->completed = true; \
-		if (rcommq_ptr->acks_recvd < spec->consistency_factor) { \
-			rcommq_ptr->status = -1; \
-			signal_luworker(); \
-		} \
-		MTX_UNLOCK(&rcommq_ptr->rcommand_mtx); \
-	} \
-}
+		while (parsed < hdr->len) {
```
vishnuitta: there is data_len in the replica_rcommon_resp_t.. not sure whether it is in use (yet to go through complete PR).. if not in use, it can be removed

src/replication.c
```diff
@@ -497,292 +180,72 @@ perform_read_write_on_fd(int fd, uint8_t *data, uint64_t len, int *errorno,
...
 
-	//below lock/unlock sequence is to make sure atomically that
-	//copies_sent is incremented by adding command to all replicas.
-	//this is actually required only in the case of
-	//(rcommq_ptr->acks_recvd + rcommq_ptr->ios_aborted == rcommq_ptr->copies_sent)
-	MTX_LOCK(&spec->rq_mtx);
-	MTX_UNLOCK(&spec->rq_mtx);
+			io_chunk_entry = malloc(sizeof(io_data_chunk_t));
+			io_chunk_entry->io_num = io_hdr->io_num;
+			io_chunk_entry->data = dataptr + sizeof(struct zvol_io_rw_hdr) + len;
```
vishnuitta: '+ len' at the end is misleading. it need to be removed.

src/replication.c
```diff
@@ -497,292 +180,72 @@ perform_read_write_on_fd(int fd, uint8_t *data, uint64_t len, int *errorno,
...
+	uint64_t parsed = 0;
+	uint8_t *read_data;
+	io_data_chunk_t *io_chunk_entry, *io_chunk_next;
 
-				revent.fd = events[i].data.fd;
-				revent.state = &(replica->io_state);
-				revent.io_hdr = replica->io_resp_hdr;
-				revent.io_data = &(replica->io_resp_data);
-				revent.byte_count = &(replica->io_read);
+	read_data = malloc(len);
```
vishnuitta: we should be using 'xmalloc'.. check with @payes once

src/data_conn.c
```diff
@@ -0,0 +1,612 @@
...
+
+		rcomm_cmd->resp_list[idx].io_resp_hdr = *(r->io_resp_hdr);
+		rcomm_cmd->resp_list[idx].data_ptr = r->ongoing_io_buf;
+		if (rcomm_cmd->opcode == ZVOL_OPCODE_WRITE)
+			rcomm_cmd->resp_list[idx].data_len = rcomm_cmd->data_len;
+		else
+			rcomm_cmd->resp_list[idx].data_len = r->ongoing_io_len;
+		rcomm_cmd->resp_list[idx].status = 1;
+
+		MTX_LOCK(mtx);
```
vishnuitta: as we already taking the lock, lets move the lock few statements above to update resp_list also
also, add comment to do atomic fetch and updates in case of removing the locks

vishnuitta: same with error case also

src/istgt_lu_disk.c
```diff
@@ -5552,68 +5564,214 @@ istgt_lu_disk_scsi_reserve(ISTGT_LU_DISK *spec, CONN_Ptr conn, ISTGT_LU_CMD_Ptr
...
+		} else {
+			data = handle_read_consistency(rcomm_cmd);
+			cmd->data = data;
+			cmd->data_len = rcomm_cmd->data_len;
+			rc = 1;
+		}
+	} else if (rcomm_cmd->opcode == ZVOL_OPCODE_WRITE) {
+		if (success >= spec->consistency_factor) {
+			rc = 1;
+			rcomm_cmd->status = 2;
```
vishnuitta: enum's are better for all these numbers

src/istgt_lu_disk.c
```diff
@@ -5552,68 +5564,214 @@ istgt_lu_disk_scsi_reserve(ISTGT_LU_DISK *spec, CONN_Ptr conn, ISTGT_LU_CMD_Ptr
...
+	// now wait for command to complete
+	while (1) {
+		MTX_LOCK(&spec->luworker_rmutex[cmd->luworkerindx]);
+		// check for status of rcomm_cmd
+		if (check_for_command_completion(spec, rcomm_cmd, cmd)) {
+			if (rcomm_cmd->status == -1)
+				rc = -1;
+			else
+				rc = cmd->data_len = rcomm_cmd->data_len;
+			put_to_mempool(&spec->rcommon_deadlist, rcomm_cmd);
```
vishnuitta: hope you taken care of not destroying rmutex and rcond variables when luworker is destroyed.

src/istgt_lu_disk.c
```diff
@@ -5552,68 +5564,214 @@ istgt_lu_disk_scsi_reserve(ISTGT_LU_DISK *spec, CONN_Ptr conn, ISTGT_LU_CMD_Ptr
...
+
+			MTX_LOCK(&spec->rq_mtx);
+			TAILQ_REMOVE(&spec->rcommon_waitq, rcomm_cmd, wait_cmd_next);
```
vishnuitta: I don't think we need luworker_rmutex lock to remove from this list. If so, luworker_rmutex can be unlocked before this

src/istgt_lu_disk.c
```diff
@@ -5552,68 +5564,214 @@ istgt_lu_disk_scsi_reserve(ISTGT_LU_DISK *spec, CONN_Ptr conn, ISTGT_LU_CMD_Ptr
...
+			MTX_UNLOCK(&spec->rq_mtx);
+
+			MTX_UNLOCK(&spec->luworker_rmutex[cmd->luworkerindx]);
+			break;
+		}
+		now = time(NULL);
+		abstime.tv_sec = now + 1;
+		abstime.tv_nsec = 0;
+
+		pthread_cond_timedwait(&spec->luworker_rcond[cmd->luworkerindx], &spec->luworker_rmutex[cmd->luworkerindx],
```
vishnuitta: I don't think we miss any signals now as lock is taken while signaling..
however, just add in comments the work that need to be done to remove locking here

src/replication.c
```diff
@@ -850,112 +322,83 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
...
-		 */
-		return NULL;
+	if (read(replica->iofd, rio, sizeof (*rio)) != sizeof (*rio)) {
+		REPLICA_ERRLOG("failed to read handshake response from replica\n");
+		goto replica_error;
 	}
 
-	return replica;
-}
+	if(init_mempool(&replica->cmdq, rcmd_mempool_count, sizeof (rcmd_t), 0,
```
vishnuitta: It doesn't seem to be having MP, SC setting for this mempool.
Please verify the flags for other mempools also.

src/replication.c
```diff
@@ -1673,15 +1197,14 @@ initialize_volume(spec_t *spec) {
...
-//When all the replicas are up, make replicas as RW and change state of istgt to REAL;
-
 
 int
-initialize_replication_mempool(bool should_fail) {
+initialize_replication_mempool(bool should_fail)
+{
 	int rc = 0;
 
 	rc = init_mempool(&rcmd_mempool, rcmd_mempool_count, sizeof (rcmd_t), 0,
```
vishnuitta: lets set rcmd_mempool_count as 5 times of rcommon_cmd_mempool_count

src/replication.c
```diff
@@ -1251,7 +711,7 @@ write_io_data(replica_t *replica, io_event_t *wevent)
 			*state = READ_IO_RESP_HDR;
 			break;
 	}
-
+	REPLICA_ERRLOG("got doneocount : %d : r:%p\n", donecount, replica);
```
vishnuitta: this log should be removed or should be moved under debug build

src/replication.c
```diff
@@ -1380,14 +839,153 @@ handle_write_data_event(replica_t *replica)
 	return rc;
 }
 
+static void
```
vishnuitta: caller seems to be taking same lock and released before calling this function.
So, this can be merged without doing lock/unlock sequence twice.

src/replication.c
```diff
@@ -1380,14 +839,153 @@ handle_write_data_event(replica_t *replica)
...
+	uint64_t num = 1;
+	r->disconnect_conn = 1;
+	if (write(r->mgmt_eventfd2, &num, sizeof (num)) != sizeof (num))
+		REPLICA_NOTICELOG("Failed to inform err to data_conn for replica(%p)\n", r);
+}
+
+static void
+free_replica(replica_t *r)
+{
+	/* TOOD: when to remove from rwaitq of spec */
```
vishnuitta: TODO can be removed

src/replication.c
```diff
@@ -1380,14 +839,153 @@ handle_write_data_event(replica_t *replica)
...
+		//ASSERT(r->conn_closed == 1);
+		/* case where error happened while sending HANDSHAKE or sending is successful but error from zvol_handshake */
+		if ((r->iofd == -1) || (r->mgmt_eventfd2 == -1)) {
+			TAILQ_FOREACH(r_ev, &(r->spec->rwaitq), r_waitnext) {
+				if (r_ev == r) {
+					TAILQ_REMOVE(&r->spec->rwaitq, r, r_waitnext);
+					r->conn_closed++;
+				}
+			}
+		} else {
```
vishnuitta: Informing replica thread to close data connection should be fine as long as mgmt_eventfd2 is not -1. For this, replica thread should set mgmt_eventfd2 to -1 with lock. Below is code change:
```diff --git a/src/data_conn.c b/src/data_conn.c
index 11b2f1e..8ae1b8f 100644
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -192,7 +192,9 @@ handle_data_conn_error(replica_t *r)
 
        MTX_UNLOCK(&spec->rq_mtx);
 
+       MTX_LOCK(&r->r_mtx);
        if (epoll_ctl(r->epollfd, EPOLL_CTL_DEL, r->iofd, NULL) == -1) {
+               MTX_UNLOCK(&r->r_mtx);
                perror("epoll_ctl");
                return -1;
        }
@@ -200,16 +202,17 @@ handle_data_conn_error(replica_t *r)
        fd = r->iofd;
        r->iofd = -1;
 
-        if (r->io_resp_hdr)
-                free(r->io_resp_hdr);
-        if (r->ongoing_io_buf)
-                free(r->ongoing_io_buf);
-
        close(fd);
 
        data_eventfd = r->data_eventfd;
        r->data_eventfd = -1;
        close_fd(epollfd, data_eventfd);
+       MTX_UNLOCK(&r->r_mtx);
+
+       if (r->io_resp_hdr)
+               free(r->io_resp_hdr);
+       if (r->ongoing_io_buf)
+               free(r->ongoing_io_buf);
 
        respond_with_error_for_all_outstanding_ios(r);
 
@@ -225,6 +228,7 @@ handle_data_conn_error(replica_t *r)
         * and mgmt_conn thread, which are consumers of mgmt_eventfd2
         */
                pthread_cond_signal(&r->r_cond);
+       r->mgmt_eventfd2 = -1;
        MTX_UNLOCK(&r->r_mtx);
 
        /* replica might have got destroyed */
diff --git a/src/replication.c b/src/replication.c
index bf780c2..61d9461 100644
--- a/src/replication.c
+++ b/src/replication.c
@@ -897,7 +897,8 @@ handle_mgmt_conn_error(replica_t *r, int sfd, struct epoll_event *events, int ev
        r->conn_closed++;
        if (r->conn_closed != 2) {
                //ASSERT(r->conn_closed == 1);
-               /* case where error happened while sending HANDSHAKE or sending is successful but error from zvol_handshake */
+               /* case where error happened while sending HANDSHAKE or sending is successful but error from zvol_handshake or
+                * data connection is closed before this */
                if ((r->iofd == -1) || (r->mgmt_eventfd2 == -1)) {
                        TAILQ_FOREACH(r_ev, &(r->spec->rwaitq), r_waitnext) {
                                if (r_ev == r) {
@@ -905,7 +906,8 @@ handle_mgmt_conn_error(replica_t *r, int sfd, struct epoll_event *events, int ev
                                        r->conn_closed++;
                                }
                        }
-               } else {
+               }
+               if (r->mgmt_eventfd2 != -1) {
                        inform_data_conn(r);
                }
        } else {
```

src/replication.c
```diff
@@ -1741,3 +1265,124 @@ destroy_relication_mempool(void) {
...
+			rcomm_cmd = get_from_mempool(&spec->rcommon_deadlist);
+
+			for (i = 0; i < rcomm_cmd->copies_sent; i++) {
+				if (rcomm_cmd->resp_list[i].status)
+					count++;
+			}
+
+			if (count == rcomm_cmd->copies_sent) {
+				destroy_resp_list(rcomm_cmd);
+				free(rcomm_cmd->iov[0].iov_base);
```
vishnuitta: there is nothing in iov[0] of rcomm_cmd.. this 'free' should not be there

src/replication.c
```diff
@@ -1741,3 +1265,124 @@ destroy_relication_mempool(void) {
...
+
+				for (i=1; i<rcomm_cmd->iovcnt + 1; i++)
```
vishnuitta: caller of 'replicate', i.e., 'istgt_lu_disk_lbwrite' seems to be freeing this iov in error cases..
those frees in 'istgt_lu_disk_lbwrite' need to be avoided for REPLICATION macro

PR #20, opened 2018-05-25
Handshake on data_io connection need to be as per uZFS requirement.

satbirchhikara:
Implemented ZVOL_OPCODE_OPEN at iSCSI controller side. Now iSCSI controller will share timeout & block size info with uZFS at time of data_io connection handshake.
Unit test: Modified existing UT case to send ZVOL_OPCODE_OPEN with required info.


satbirchhikara:
Closing this pull request as I have created new one with upstream sync and more changes.

PR #23, opened 2018-05-26
Changed `size_t` to `ssize_t` because `size_t` is unsigned integer type.

shubhambhattar:
Value of `size_t` cannot be negative because it is an unsigned integer type. Thus, `len <= 0` when `len` is `size_t` type makes no sense (except the case when `len == 0`). Changed it to `int` instead.

Changes made in `istgt_md5.c` (function definition) and `istgt_md5.h` (function prototype declaration) files.

### DIFF
--- a/src/istgt_md5.c
+++ b/src/istgt_md5.c
@@ -61,7 +61,7 @@ istgt_md5final(void *md5, ISTGT_MD5CTX *md5ctx)
 }
 
 int
-istgt_md5update(ISTGT_MD5CTX *md5ctx, const void *data, size_t len)
+istgt_md5update(ISTGT_MD5CTX *md5ctx, const void *data, int len)
 {
 	int rc;
 
--- a/src/istgt_md5.c
+++ b/src/istgt_md5.c
@@ -61,7 +61,7 @@ istgt_md5final(void *md5, ISTGT_MD5CTX *md5ctx)
 }
 
 int
-istgt_md5update(ISTGT_MD5CTX *md5ctx, const void *data, int len)
+istgt_md5update(ISTGT_MD5CTX *md5ctx, const void *data, ssize_t len)
 {
 	int rc;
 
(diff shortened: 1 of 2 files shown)


shubhambhattar:
@satyamz Can you help me with this? I don't know why the build is failing. I found some errors and wanted to make a pull request for the same.

shubhambhattar:
I did not test these changes in my setup. I just made the required changes and waited for Travis check. The first time, I missed to change the function prototype declaration in the header file ([#21](https://github.com/openebs/istgt/pull/21)). But the build is failing even after correcting it. Can you help me on what might go wrong or how to pin point the actual problem?

src/istgt_md5.c
```diff
@@ -61,7 +61,7 @@ istgt_md5final(void *md5, ISTGT_MD5CTX *md5ctx)
 }
 
 int
-istgt_md5update(ISTGT_MD5CTX *md5ctx, const void *data, size_t len)
+istgt_md5update(ISTGT_MD5CTX *md5ctx, const void *data, int len)
```
mynktl: If you are planning to change `size_t` to `int` then change it to `ssize_t` as `size_t` is `unsigned long int` and `ssize_t` is `signed long int`. Other way is to change the condition `len <= 0` to` len == 0`.

mynktl:
@shubhambhattar  Thanks for changes. Travis build is set to run on replication branch, not on the master branch. Can you change the destination branch to replication instead of master?

src/istgt_md5.c
shubhambhattar: I didn't change the condition because I didn't know if negative `len` can actually occur or not. This is why I tried to change the data type.

shubhambhattar:
@mynktl Done. Pull request is done for replication branch.

src/istgt_md5.c
vishnuitta: can you please change size_t to ssize_t rather than int?

shubhambhattar: Okay. Will do.

shubhambhattar: Done.

vishnuitta:
This PR has been asked to raise against 'replication' branch, which have the travis integrated.
Closing this PR which is against 'master' branch, and, PR #27 is raised against 'replication' branch.

PR #25, opened 2018-05-28
Added functionality in iSCSI controller to own dataset on data-io con…

satbirchhikara:
Implemented ZVOL_OPCODE_OPEN at iSCSI controller side. Now iSCSI controller will share timeout & block size info with uZFS at time of data_io connection handshake.
Unit test: Modified existing UT case to send ZVOL_OPCODE_OPEN with required info.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -12,6 +12,7 @@
 #include <errno.h>
 #include <sys/prctl.h>
 #include <sys/eventfd.h>
+#include </usr/include/assert.h>
 #include "zrepl_prot.h"
 #include "replication.h"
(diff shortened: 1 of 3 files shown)


src/replication.c
```diff
@@ -12,6 +12,7 @@
 #include <errno.h>
 #include <sys/prctl.h>
 #include <sys/eventfd.h>
+#include </usr/include/assert.h>
```
vishnuitta: having relative path is better

satbirchhikara: Since no assert is included at this point of time, I removed inclusion of file as per your comments.

PR #27, opened 2018-05-28
Changed `size_t` to `int` because `size_t` is unsigned integer type

shubhambhattar:
Value of `size_t` cannot be negative because it is an unsigned integer type. Thus, `len <= 0` when `len` is `size_t` type makes no sense (except the case when `len == 0`). Changed it to `int` instead.

Changes made in `istgt_md5.c` (function definition) and `istgt_md5.h` (function prototype declaration) files.

### DIFF
--- a/src/istgt_md5.c
+++ b/src/istgt_md5.c
@@ -61,7 +61,7 @@ istgt_md5final(void *md5, ISTGT_MD5CTX *md5ctx)
 }
 
 int
-istgt_md5update(ISTGT_MD5CTX *md5ctx, const void *data, size_t len)
+istgt_md5update(ISTGT_MD5CTX *md5ctx, const void *data, int len)
 {
 	int rc;
 
--- a/src/istgt_md5.c
+++ b/src/istgt_md5.c
@@ -61,7 +61,7 @@ istgt_md5final(void *md5, ISTGT_MD5CTX *md5ctx)
 }
 
 int
-istgt_md5update(ISTGT_MD5CTX *md5ctx, const void *data, int len)
+istgt_md5update(ISTGT_MD5CTX *md5ctx, const void *data, ssize_t len)
 {
 	int rc;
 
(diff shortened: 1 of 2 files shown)


vishnuitta:
@shubhambhattar .. travis failed with this commit.. so, I reverted the commit.
Coming to the fix,
- let us leave this as 'size_t'
- make the len check only for '0'
- move the negative check to callers of this function

shubhambhattar:
@vishnuitta Okay, so I have a doubt. Let's take the example case in file `src/istgt_iscsi.c`. We have the following lines:
```
istgt_md5update(&md5ctx, conn->auth.chap_id, 1);   // Call 1
/* followed by secret */
istgt_md5update(&md5ctx, conn->auth.secret,
    strlen(conn->auth.secret));                                      // Call 2
/* followed by Challenge Value */
istgt_md5update(&md5ctx, conn->auth.chap_challenge,
    conn->auth.chap_challenge_len);                          // Call 3
```

- I don't need to check anything for Call 1 as I am explicitly passing the value `1` which will be correctly captured by `size_t`. 
- Similarly I don't need to check anything for Call 2 as return value of `strlen` will be `>= 0` which will be correctly captured by `size_t`
- Now, for call 3, I should logically put a check something like:

```
if(conn->auth.chap_challenge_len < 0)
    // do something
```

What should I write in the `// do something` part? Similar is the case in some other files where the function is called. I'll put those checks in all the files.

Also, the `istgt_md5update()` function declaration states that it returns a value of type `int` but the value is not captured in any of the places where this function is called. Why so? Or am I missing something?

mynktl:
Hey @shubhambhattar,

Please put a condition in caller as you mentioned in your previous message.. like,
`if(conn->auth.chap_challenge_len > 0)`
           `istgt_md5update(); `

This way we can avoid updating function definiation.

shubhambhattar:
@mynktl Check [#32](https://github.com/openebs/istgt/pull/32). I have made the changes as per your instructions.

Also, please add the `summerhack` label.

PR #28, opened 2018-05-28
TA372 :  replica timeout handling

mynktl:
Changes : 
- Handling of replica timeout. If wait time for IO completion from replica increases beyond specified timeout then replica will get disconnected from ISTGT.
- Added new arguments to istgt . (`-R value`, where the value is number of seconds to wait for replica response)
- Removing lock protection while updating response from replica to ISTGT
- Updating timeout in replicate function to 500ms instead of 1 seconds.

Changes in replication_test: 
- Changes related to handling of arguments
`replication_test -i target_ip -p target_port -I replica_ip -P replica_port -V volume_path -t io_cnt`
If `-t 30` argument is specified then replication_test will server first 30 IOs(read or write) then sleep for 60 seconds to causes a timeout in ISTGT. (60 seconds is an experimental value which can be changed in code according to requirement)


src/data_conn.c
```diff
@@ -600,13 +603,21 @@ replica_thread(void *arg)
 			if (ret == -1)
 				goto exit;
 		}
-		/* Need to handle wait IOs that are taking more than IO timeout */
-		// handle_data_conn_error(r);
-		// return -1;
+
+		/* check for replica IO timeout */
+		pending_cmd = TAILQ_FIRST(&r->waitq);
```
vishnuitta: we need to be looking at HEAD of all queues..

src/data_conn.c
```diff
@@ -247,6 +250,7 @@ handle_epoll_out_event(replica_t *r)
 			return -1;
 		if (ret == WRITE_COMPLETED) {
 			TAILQ_REMOVE(&r->readyq, cmd, next);
+			clock_gettime(CLOCK_MONOTONIC, &cmd->queued_time);
```
vishnuitta: let us set the queued time from the point when it gets added to blocked q or ready q

src/data_conn.c
```diff
@@ -600,13 +603,21 @@ replica_thread(void *arg)
...
 		}
-		/* Need to handle wait IOs that are taking more than IO timeout */
-		// handle_data_conn_error(r);
-		// return -1;
+
+		/* check for replica IO timeout */
+		pending_cmd = TAILQ_FIRST(&r->waitq);
+		if (pending_cmd != NULL) {
+			clock_gettime(CLOCK_MONOTONIC, &now);
+			timesdiff(pending_cmd->queued_time, now, diff);
```
vishnuitta: let us change epoll_timeout based on when the next iteration should be, so that, an IO doesn't hang for time longer than the configured replica timeout

src/data_conn.c
```diff
@@ -30,6 +31,8 @@ int handle_epoll_in_event(replica_t *r);
 static void inform_mgmt_conn(replica_t *r);
 static rcmd_t *dequeue_replica_cmdq(replica_t *replica);
 
+int replica_timeout = REPLICA_DEFAULT_TIMEOUT;
```
vishnuitta: description here

src/istgt.c
```diff
@@ -2580,7 +2581,8 @@ usage(void)
 	printf(" -D         don't detach from tty\n");
 	printf(" -H         show this usage\n");
 	printf(" -V         show version\n");
-	printf(" -P         Persist Disabled\n");	
+	printf(" -P         Persist Disabled\n");
+	printf(" -R         Replica timeout in seconds\n");
```
vishnuitta: 'IO timeout in seconds at replicas'

src/replication_test.c
```diff
@@ -197,48 +192,94 @@ send_io_resp(int fd, zvol_io_hdr_t *io_hdr, void *buf)
...
-	
-	data = NULL;
+	zvol_io_hdr_t *io_hdr;
+	zvol_io_hdr_t *mgmtio;
 	bool read_rem_data = false;
 	bool read_rem_hdr = false;
 	uint64_t recv_len = 0;
 	uint64_t total_len = 0;
 	uint64_t io_hdr_len = sizeof(zvol_io_hdr_t);
+	int io_timeout = 0;
```
vishnuitta: instead of io_timeout, make it as io_cnt

src/replication_test.c
```diff
@@ -197,48 +192,94 @@ send_io_resp(int fd, zvol_io_hdr_t *io_hdr, void *buf)
...
+	int ch;
+	int check = 1;
+
+	while ((ch = getopt(argc, argv, "i:p:I:P:V:t:")) != -1) {
+		switch (ch) {
+			case 'i':
+				strncpy(ctrl_ip, optarg, sizeof(ctrl_ip));
+				check = 1 << 1;
```
vishnuitta: you should be doing '|' here, and checking for check should be changed accordingly..

src/data_conn.c
mynktl: ok

src/data_conn.c
mynktl: ok

src/data_conn.c
mynktl: ok

src/data_conn.c
mynktl: ok

src/istgt.c
mynktl: ok

src/replication_test.c
mynktl: ok

src/replication_test.c
mynktl: ok

src/replication.h
```diff
@@ -154,6 +155,9 @@ extern void * replica_thread(void *);
 extern int do_drainfd(int );
 void close_fd(int epollfd, int fd);
 
+/* Replica default timeout is 60 seconds */
+#define	REPLICA_DEFAULT_TIMEOUT	60
```
vishnuitta: set this to 200

src/data_conn.c
```diff
@@ -609,13 +653,14 @@ replica_thread(void *arg)
 			if (ret == -1)
 				goto exit;
 		}
-		/* Need to handle wait IOs that are taking more than IO timeout */
-		// handle_data_conn_error(r);
-		// return -1;
+
+		/* check for replica IO timeout */
+		CHECK_REPLICA_TIMEOUT(&r->readyq, diff_wq, ret, exit, epoll_timeout);
```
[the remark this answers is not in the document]
mynktl: ok

src/replication.h
mynktl: ok

src/data_conn.c
```diff
@@ -76,6 +80,31 @@ static rcmd_t *dequeue_replica_cmdq(replica_t *replica);
...
+			clock_gettime(CLOCK_MONOTONIC, &now);		\
+			timesdiff(pending_cmd->queued_time, now, _diff);\
+			if (_diff.tv_sec >= replica_timeout) {		\
+				REPLICA_ERRLOG("timeout happened for "	\
+				    "replica(%s:%d)..time(%lu)\n", 	\
+				    r->ip, r->port, _diff.tv_sec); 	\
+				_ret = -1;				\
+				goto _exit_label;			\
+			} else {					\
+				ms = _diff.tv_sec * 1000;		\
```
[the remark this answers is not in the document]
mynktl: we can avoid it.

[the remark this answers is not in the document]
mynktl: it will not be executed per IO.

PR #29, opened 2018-05-29
Revert "Changed `size_t` to `int` because `size_t` is unsigned integer type"

vishnuitta:
Reverts openebs/istgt#27
there is compilation issue with this change. Hence, reverting the change.


shubhambhattar:
Can you add the `summerhack` label here?

PR #31, opened 2018-05-30
TA343 [DE18 : marking replica as down in target]

mynktl:
Changes : 
- Added comments in replica related code
- Used `cmd_state_t` in place of static number for rcommon_cmd state

### DIFF
--- a/src/istgt_lu_disk.c
+++ b/src/istgt_lu_disk.c
@@ -104,6 +104,10 @@ extern rte_smempool_t rcommon_cmd_mempool;
 		    &spec->luworker_rcond[cmd->luworkerindx];		\
 		rcomm_cmd->healthy_count = spec->healthy_rcount;	\
 		rcomm_cmd->io_seq = ++spec->io_seq;			\
+		rcomm_cmd->replication_factor = 			\
+		    spec->replication_factor;				\
+		rcomm_cmd->consistency_factor =				\
+		    spec->consistency_factor;				\
 		rcomm_cmd->state = CMD_ENQUEUED_TO_WAITQ;		\
 		switch (cmd->cdb0) {					\
 			case SBC_WRITE_6:				\
@@ -5661,7 +5665,7 @@ check_for_command_completion(spec_t *spec, rcommon_cmd_t *rcomm_cmd, ISTGT_LU_CM
 			rc = 1;
 		}
 	} else if (rcomm_cmd->opcode == ZVOL_OPCODE_WRITE) {
-		if (success >= spec->consistency_factor) {
+		if (success >= rcomm_cmd->consistency_factor) {
 			rc = 1;
 		} else if ((success + failure) == rcomm_cmd->copies_sent) {
 			rc = -1;
@@ -5683,7 +5687,7 @@ replicate(ISTGT_LU_DISK *spec, ISTGT_LU_CMD_Ptr cmd, uint64_t offset, uint64_t n
 	int iovcnt = cmd->iobufindx + 1;
 	bool cmd_sent = false;
 	struct timespec abstime, now;
-	int nsec;
+	int nsec, err_num = 0;
 
 	MTX_LOCK(&spec->rq_mtx);
 	if(spec->ready == false) {
@@ -5740,7 +5744,7 @@ replicate(ISTGT_LU_DISK *spec, ISTGT_LU_CMD_Ptr cmd, uint64_t offset, uint64_t n
 			/*
 			 * NOTE: This is for debugging purpose only
 			 */
-			if (errno == ETIMEDOUT)
+			if (err_num == ETIMEDOUT)
 				fprintf(stderr,"last errno(%d) opcode(%d)\n",
 				    errno, rcomm_cmd->opcode);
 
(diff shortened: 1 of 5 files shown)


src/istgt_lu_disk.c
```diff
@@ -5668,17 +5658,13 @@ check_for_command_completion(spec_t *spec, rcommon_cmd_t *rcomm_cmd, ISTGT_LU_CM
 		} else {
 			data = handle_read_consistency(rcomm_cmd);
 			cmd->data = data;
-			cmd->data_len = rcomm_cmd->data_len;
 			rc = 1;
 		}
 	} else if (rcomm_cmd->opcode == ZVOL_OPCODE_WRITE) {
 		if (success >= spec->consistency_factor) {
```
vishnuitta: let us store consistency_factor also in rcomm_cmd, so that, any change in it won't be the problem.

src/istgt_lu_disk.c
```diff
@@ -5668,17 +5658,13 @@ check_for_command_completion(spec_t *spec, rcommon_cmd_t *rcomm_cmd, ISTGT_LU_CM
...
 			rc = 1;
-			rcomm_cmd->status = 2;
```
vishnuitta: why we removed this line? and also the change below this?

src/istgt_lu_disk.c
```diff
@@ -5668,17 +5658,13 @@ check_for_command_completion(spec_t *spec, rcommon_cmd_t *rcomm_cmd, ISTGT_LU_CM
 		} else {
```
vishnuitta, line 5662: this 'else' is not correct as explained in previous PR.
Lets say, copies_sent is 10, success is 1, failure is 9. In this case, success response is sent, which is not correct.
We atleast need (MAX(n-m+1, m)) number of success responses from degraded replicas to start building the read response (or) success response from one healthy replica is required.

src/istgt_lu_disk.c
```diff
@@ -5749,18 +5729,24 @@ replicate(ISTGT_LU_DISK *spec, ISTGT_LU_CMD_Ptr cmd, uint64_t offset, uint64_t n
...
+			if (rc == 1)
 				rc = cmd->data_len = rcomm_cmd->data_len;
+			rcomm_cmd->state = CMD_EXECUTION_DONE;
 			put_to_mempool(&spec->rcommon_deadlist, rcomm_cmd);
+			MTX_UNLOCK(rcomm_cmd->mutex);
+
+			/*
+			 * NOTE: This is for debugging purpose only
+			 */
+			if (errno == ETIMEDOUT)
```
vishnuitta: 'errno' is supposed to be of pthread_cond_timedwait. But, by the time execution reaches here, 'errno' might have got overwritten by other syscalls.
So, we need to store 'errno' into local variable.

src/istgt_lu_disk.c
vishnuitta: ok.. status is removed completely..

src/replication.c
```diff
@@ -110,11 +111,11 @@ update_volstate(spec_t *spec)
  * and sets fd_closed also closes fd if read return 0, i.e., EOF
  * returns number of bytes read/written
  */
-static int64_t
+ssize_t
```
vishnuitta, line 114: should we change the caller functions variables also?

src/replication.c
```diff
@@ -151,14 +152,18 @@ perform_read_write_on_fd(int fd, uint8_t *data, uint64_t len, int state)
...
 
 	return nbytes;
 }
 
+/*
+ * get_all_read_resp_data_chunk will create/update io_data_chunk_list
+ * with response received from replica according to io_numer.
+ */
 void
 get_all_read_resp_data_chunk(replica_rcomm_resp_t *resp, struct io_data_chunk_list_t *io_chunk_list)
```
vishnuitta, line 168: this needs fix in case of all degraded replicas scenario. if I'm not wrong, this will be handled in next PRs.

src/replication.c
```diff
@@ -1426,7 +1473,7 @@ initialize_volume(spec_t *spec)
 	TAILQ_INIT(&spec->rq);
 	TAILQ_INIT(&spec->rwaitq);
 
-        if(init_mempool(&spec->rcommon_deadlist, rcmd_mempool_count, sizeof (rcmd_t), 0,
+        if(init_mempool(&spec->rcommon_deadlist, rcmd_mempool_count, 0, 0,
```
vishnuitta, line 1477: default seems to be SP/SC for ring.. it need to be changed according to the ring usage (for all rings)

vishnuitta: ok.. we are using MP/MC.. its fine then

src/istgt_lu_disk.c
mynktl: I haven't added read consistency related changes in this PR.
I was thinking of handling this changes in US1320 PR.

src/istgt_lu_disk.c
mynktl: consistency_factor is protected through spec level mutex. If we store consistency_factor in rcomm_cmd, we won't be able to reflect the change in consistency factor to old queued command.

src/istgt_lu_disk.c
mynktl: ok

src/replication.c
mynktl: All Caller function variables are signed integer or signed long int. I will change signed int to ssize_t.

src/replication.c
mynktl: yes

PR #33, opened 2018-06-01
integration testing framework for istgt

vishnuitta:
This PR is to have a integration testing framework for replication / rebuilding feature in istgt.
Approach followed is:
- mocking replica to handle read/write and other management commands
- mocking client to send read / write requests to replication layer in istgt
- creating a separate binary, so that, testing code will not get into production binary
- integrate with travis

### DIFF
--- a/src/istgt_integration_test.c
+++ b/src/istgt_integration_test.c
@@ -8,9 +8,14 @@
 #include <stdlib.h>
 #include <unistd.h>
 
+#include "istgt_misc.h"
+#include "istgt_proto.h"
 #include "replication.h"
 #include "istgt_integration.h"
 #include "replication_misc.h"
+
+__thread char tinfo[20] = {0};
+int g_trace_flag = 0;
 
 typedef struct rargs_s {
 	char replica_ip[MAX_IP_LEN];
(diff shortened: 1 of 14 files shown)


vishnuitta:
@satbirchhikara @payes @mynktl 
This PR is ready for your review. I will do conflicts resolution once other PR from Mayank gets merged.
Thanks.

src/istgt_integration_test.c
```diff
@@ -0,0 +1,726 @@
...
+	if (rc != 0) {
+		REPLICA_ERRLOG("error in initializing spec..\n");
+		return 1;
+	}
+
+	initialize_volume(spec, replication_factor, consistency_factor);
+
+	pthread_create(&replica_thread, NULL, &init_replication, (void *)NULL);
+
+	spec->ready = false;
```
satbirchhikara, line 829: I think it is not required as we already set it to false in initialize_volume.

src/istgt_integration_test.c
```diff
@@ -0,0 +1,726 @@
...
+	pthread_mutex_init(&rargs->io_recv_mtx, NULL);
+	pthread_mutex_init(&rargs->io_send_mtx, NULL);
+
+	pthread_cond_init(&rargs->io_recv_cv, NULL);
+	pthread_cond_init(&rargs->io_send_cv, NULL);
+
+	TAILQ_INIT(&rargs->io_recv_list);
+	TAILQ_INIT(&rargs->io_send_list);
+
+	pthread_create(&mgmt_receiver, NULL, &mock_repl_mgmt_receiver, args);
```
satbirchhikara, line 626: Precautionary, we can arrange this code such that before we create receiver thread, we should have sender/worker threads created.

src/istgt_integration_test.c
```diff
@@ -0,0 +1,726 @@
...
+	pthread_create(&mgmt_sender, NULL, &mock_repl_mgmt_sender, args);
+	pthread_create(&mgmt_worker, NULL, &mock_repl_mgmt_worker, args);
+
+	while (1) {
+		rargs->iofd = accept(sfd, &saddr, &slen);
```
satbirchhikara, line 631: Same goes here too.

PR #35, opened 2018-06-05
Handling read command when all replicas are in the degraded mode

mynktl:
Changes: 
- Handling read command when all replicas are in the degraded mode
- test added to test_istgt.sh to verify read consistency when all replicas are in the degraded mode


src/replication.c
```diff
@@ -1324,20 +1373,43 @@ check_for_command_completion(spec_t *spec, rcommon_cmd_t *rcomm_cmd, ISTGT_LU_CM
 		}
 	}
 
+	response_recieved = success + failure;
+	min_response = MAX(spec->replication_factor -
```
vishnuitta: we should be taking these from rcommon_cmd structure

src/replication.c
```diff
@@ -1314,7 +1362,8 @@ check_for_command_completion(spec_t *spec, rcommon_cmd_t *rcomm_cmd, ISTGT_LU_CM
 {
 	int i, rc = 0;
 	uint8_t *data = NULL;
-	int success = 0, failure = 0;
+	int success = 0, failure = 0, response_recieved;
```
vishnuitta: spelling mistake 'response_received'

src/replication.c
```diff
@@ -218,41 +218,46 @@ perform_read_write_on_fd(int fd, uint8_t *data, uint64_t len, int state)
...
 		while (parsed < hdr->len) {
 			io_hdr = (struct zvol_io_rw_hdr *)dataptr;
-
-			if (io_chunk_entry->io_num < io_hdr->io_num) {
-				io_chunk_entry->data = dataptr + sizeof(struct zvol_io_rw_hdr) + len;
-				io_chunk_entry->io_num = io_hdr->io_num;
-				io_chunk_entry->len = io_hdr->len;
+			data_len = 0;
+			while (data_len < io_hdr->len) {
+				if (io_chunk_entry->io_num < io_hdr->io_num) {
```
vishnuitta, line 239: there can be case where io_num is 0, as replica turned healthy by the time IO reached it, but, target found it to be in degraded when it sent the IO.

src/replication.c

src/replication.c

src/replication.c
```diff
@@ -1324,20 +1382,43 @@ check_for_command_completion(spec_t *spec, rcommon_cmd_t *rcomm_cmd, ISTGT_LU_CM
...
+					    true);
+					cmd->data = data;
+					rc = 1;
+				}
+			}
 		} else {
-			data = handle_read_consistency(rcomm_cmd);
-			cmd->data = data;
-			rc = 1;
+			if (rcomm_cmd->copies_sent == success) {
```
vishnuitta: add comment that copies_sent should be 1 in this case

src/replication.c
```diff
@@ -1324,20 +1382,43 @@ check_for_command_completion(spec_t *spec, rcommon_cmd_t *rcomm_cmd, ISTGT_LU_CM
...
+
 	if (rcomm_cmd->opcode == ZVOL_OPCODE_READ) {
-		if ((rcomm_cmd->copies_sent != (success + failure))) {
-			rc = 0;
-		} else if (rcomm_cmd->copies_sent == failure) {
-			rc = -1;
+		response_received = success + failure;
+
+		if (rcomm_cmd->copies_sent >= min_response) {
+			if (response_received < min_response) {
```
vishnuitta: in this case i.e., copies_sent >= min_response, just checking for success condition and failure condition should be good enough. So:
- does success received >= min_response, if so, set rc as 1 and do handle_read_consistency
- else if response_received == copies_sent, set rc as -1 (becz required number of success is not received)
- anything else, would be rc as 0, which means, we need to wait for more responses

src/replication.c
```diff
@@ -1324,20 +1382,43 @@ check_for_command_completion(spec_t *spec, rcommon_cmd_t *rcomm_cmd, ISTGT_LU_CM
 		}
 	}
 
+	response_received = success + failure;
+	min_response = MAX(rcomm_cmd->replication_factor -
```
vishnuitta: min_response also depends on number of healthy and degraded replicas in the system when the responses are received.
For read IO, success response from one healthy replica is fine.
For write IO, success response from (consistency_factor) number of healthy replicas is fine.

For read IOs, below code (with comments incorporated) should be fine (we can re-arrange code to make this simple), even though, there is check for min_response which happens only for few IOs.
For write IOs, with this check for min_response, we are making the consistency factor as MAX(RF-CF+1, CF). We need to count success responses from healthy replicas, and success responses from degraded replicas to fix it. Think of case where 7 as RF, and 3 as CF.

src/replication.c
```diff
@@ -1314,7 +1371,8 @@ check_for_command_completion(spec_t *spec, rcommon_cmd_t *rcomm_cmd, ISTGT_LU_CM
 {
 	int i, rc = 0;
 	uint8_t *data = NULL;
-	int success = 0, failure = 0;
```
vishnuitta, line 1317: Few edge cases I can think of are (for both read and write):
- CF, RF as 1 i.e., single replica case where it might be healthy or degraded
- status of replicas changes or number of replicas connected changes after IOs are sent
- similar to above thing, status of replicas changes or number of replicas connected changes after verifying for spec's ready flag. One thing to consider here is: all replicas are disconnected after verifying spec's ready flag, so, there are no replicas at all. So, copies_sent would be 0.

src/replication.c
mynktl: ok. I have added a flag for reading metadata in zvol_io_hdr. By this we can solve this issue.

src/replication.c

src/replication.c

src/replication.c
mynktl: ok. I will modify write command handling.

src/replication.c
mynktl: 1. CF, RF as 1 i.e., single replica case where it might be healthy or degraded
> *In this case, `min_response` and `copies_sent` will be 1. In read case, if `success == 1`*
*then we will reply with `rc=1` or if `response_received == rcomm_cmd->copies_sent` then*
*we will reply with `rc = -1`.*
2. status of replicas changes or number of replicas connected changes after IOs are sent
> *We are taking spec lock while assigning IOs to replica and during error case handling for*
*the replica. So, it will not create the problem in this case.*
3. status of replicas changes or number of replicas connected changes after verifying for spec's ready flag
> *same as in 2nd case*

src/replication.c
```diff
@@ -1371,57 +1373,87 @@ check_for_command_completion(spec_t *spec, rcommon_cmd_t *rcomm_cmd, ISTGT_LU_CM
...
+			 */
+			REPLICA_ERRLOG("got error from %d replicas.. io(%lu)"
+			    "cs(%d)\n", failure, rcomm_cmd->io_seq,
+			    rcomm_cmd->copies_sent);
+			rc = -1;
 		}
 	} else if (rcomm_cmd->opcode == ZVOL_OPCODE_WRITE) {
-		if (success >= min_response) {
+		if (healthy_replica >= rcomm_cmd->consistency_factor &&
+		    healthy_response >= rcomm_cmd->consistency_factor) {
```
vishnuitta: just checking for healthy_response should be good. healthy_replica check can be added as assert in debug builds.

src/data_conn.c
```diff
@@ -55,7 +55,7 @@ int replica_timeout = REPLICA_DEFAULT_TIMEOUT;
 		rcomm_cmd->resp_list[idx].io_resp_hdr.status =		\
 		    ZVOL_OP_STATUS_FAILED;				\
 		rcomm_cmd->resp_list[idx].data_ptr = NULL;		\
-		rcomm_cmd->resp_list[idx].status = 1;			\
+		rcomm_cmd->resp_list[idx].status |= RECEIVED_ERR;	\
```
vishnuitta, line 58: logs can be added for this case in debug build.

vishnuitta (approved):
changes are good.

src/replication.c
mynktl: yes

src/data_conn.c
mynktl: ok

PR #37, opened 2018-06-08
[US1299] Rebuild functionality added in iSCSI controller.

satbirchhikara:
Rebuild functionality has been added in iSCSI controller. This code suppose to handle
1. Mesh rebuild i.e. iSCSI controller failure.
2. Normal rebuild i.e. rebuild from healthy replica.

What can be expected from this PR ?
1. Code should be able to handle iSCSI controller crash i.e. electing on replica for rebuild and rebuild this replica with the help of all other replicas. Mark this replica healthy and pick next replica for rebuild.
2. Rebuild was going on a replica, rebuild failed because of different reasons. It should able to handle failure case and kick-start rebuild again.
3. New replica got added, code should be able to kick-start rebuild on this replica.
4. Replica got disconnected, code should be able to see if replication_factor and consistency factor requirement is matching or not. Based on that it mark volume healthy or degraded. That way read request on target will decide if they need to read it from all online replicas or from just one.

What is missing from this PR ?
This code is not handling the case where we have consistency factor == 1 and replication factor == 1. So in that case if there is just one replica, it should be marked healthy at uZFS side. I am creating task for this.

 

### DIFF
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -226,11 +226,11 @@ handle_data_conn_error(replica_t *r)
 	else if (r->state == ZVOL_STATUS_DEGRADED)
 		spec->degraded_rcount--;
 
-	if (spec->master_replica == r) {
-		REPLICA_ERRLOG("Replica %s %d was under rebuild,"
+	if (spec->target_replica == r) {
+		REPLICA_ERRLOG("Replica:%s port:%d was under rebuild,"
 		    " seting master_replica to NULL\n",
 		    r->ip, r->port);
-		spec->master_replica = NULL;
+		spec->target_replica = NULL;
 		spec->rebuild_in_progress = false;
 	}
 
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -225,15 +225,6 @@ handle_data_conn_error(replica_t *r)
 		spec->healthy_rcount--;
 	else if (r->state == ZVOL_STATUS_DEGRADED)
 		spec->degraded_rcount--;
-
-	if (spec->target_replica == r) {
-		REPLICA_ERRLOG("Replica:%s port:%d was under rebuild,"
-		    " seting master_replica to NULL\n",
-		    r->ip, r->port);
-		spec->target_replica = NULL;
-		spec->rebuild_in_progress = false;
-	}
-
 	update_volstate(r->spec);
 
 	mgmt_eventfd2 = r->mgmt_eventfd2;
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -226,6 +226,14 @@ handle_data_conn_error(replica_t *r)
 	else if (r->state == ZVOL_STATUS_DEGRADED)
 		spec->degraded_rcount--;
 
+	if (spec->master_replica == r) {
+		REPLICA_ERRLOG("Replica %s %d was under rebuild,"
+		    " seting master_replica to NULL\n",
+		    r->ip, r->port);
+		spec->master_replica = NULL;
+		spec->rebuild_in_progress = false;
+	}
+
 	update_volstate(r->spec);
 
 	mgmt_eventfd2 = r->mgmt_eventfd2;
@@ -641,5 +649,6 @@ replica_thread(void *arg)
 exit:
 	if (ret == -1)
 		handle_data_conn_error(r);
+	REPLICA_ERRLOG("replica_thread exiting ...\n");
 	return NULL;
 }
--- a/src/istgt_lu.h
+++ b/src/istgt_lu.h
@@ -820,12 +820,18 @@ typedef struct istgt_lu_disk_t {
 	rte_smempool_t rcommon_deadlist;	// Contains completed IOs
 	TAILQ_HEAD(, replica_s) rq; //Queue of replicas connected to this spec(volume)
 	TAILQ_HEAD(, replica_s) rwaitq; //Queue of replicas completed handshake, and yet to have data connection to this spec(volume)
-	int replica_count;
+	int replica_io_timeout;
 	int replication_factor;
 	int consistency_factor;
 	int healthy_rcount;
 	int degraded_rcount;
 	bool ready;
+	bool rebuild_in_progress;
+	int rebuild_replica_count;
+	int rebuild_info_ack_recvd;
+	struct replica_s *master_replica;
+	mgmt_ack_t *rebuild_info;
+
 	/*Common for both the above queues,
 	Since same cmd is part of both the queues*/
 	pthread_mutex_t rq_mtx; 
--- a/src/istgt_lu.h
+++ b/src/istgt_lu.h
@@ -820,17 +820,13 @@ typedef struct istgt_lu_disk_t {
 	rte_smempool_t rcommon_deadlist;	// Contains completed IOs
 	TAILQ_HEAD(, replica_s) rq; //Queue of replicas connected to this spec(volume)
 	TAILQ_HEAD(, replica_s) rwaitq; //Queue of replicas completed handshake, and yet to have data connection to this spec(volume)
-	int replica_io_timeout;
 	int replication_factor;
 	int consistency_factor;
 	int healthy_rcount;
 	int degraded_rcount;
 	bool ready;
 	bool rebuild_in_progress;
-	int rebuild_replica_count;
-	int rebuild_info_ack_recvd;
-	struct replica_s *master_replica;
-	mgmt_ack_t *rebuild_info;
+	struct replica_s *target_replica;
 
 	/*Common for both the above queues,
 	Since same cmd is part of both the queues*/
--- a/src/replication.c
+++ b/src/replication.c
@@ -1237,9 +1237,7 @@ update_replica_status(spec_t *spec, replica_t *replica)
 		    	    " seting master_replica to NULL\n",
 		    	    replica->ip, replica->port);
 		}
-	}
-#if 0
- 	else if ((repl_status->state == ZVOL_STATUS_DEGRADED) &&
+	} else if ((repl_status->state == ZVOL_STATUS_DEGRADED) &&
 	    (repl_status->rebuild_status == ZVOL_REBUILDING_FAILED) &&
 	    (replica == spec->target_replica)) {
 		/*
@@ -1249,7 +1247,7 @@ update_replica_status(spec_t *spec, replica_t *replica)
 		spec->rebuild_in_progress = false;
 		spec->target_replica = NULL;
 	}
-#endif
+
 	/*Trigger rebuild if possible */
 	trigger_rebuild(spec);
 	MTX_UNLOCK(&spec->rq_mtx);
@@ -2008,14 +2006,6 @@ handle_mgmt_conn_error(replica_t *r, int sfd, struct epoll_event *events, int ev
 	mgmtfd = r->mgmt_fd;
 	r->mgmt_fd = -1;
 	MTX_UNLOCK(&r->r_mtx);
-
-	if (r->spec->target_replica == r) {
-		REPLICA_ERRLOG("Replica:%s port:%d was under rebuild,"
-		    " seting master_replica to NULL\n",
-		    r->ip, r->port);
-		r->spec->target_replica = NULL;
-		r->spec->rebuild_in_progress = false;
-	}
 	MTX_UNLOCK(&r->spec->rq_mtx);
 
 	close_fd(epollfd, mgmtfd);
@@ -2055,6 +2045,16 @@ handle_mgmt_conn_error(replica_t *r, int sfd, struct epoll_event *events, int ev
 				REPLICA_ERRLOG("unexpected fd(%d) for replica:%p\n", mevent->fd, r);
 		}
 	}
+
+	MTX_LOCK(&r->spec->rq_mtx);
+	if (r->spec->target_replica == r) {
+		REPLICA_ERRLOG("Replica:%s port:%d was under rebuild,"
+		    " seting master_replica to NULL\n",
+		    r->ip, r->port);
+		r->spec->target_replica = NULL;
+		r->spec->rebuild_in_progress = false;
+	}
+	MTX_UNLOCK(&r->spec->rq_mtx);
 
 	if (r->dont_free != 1)
 		free_replica(r);
--- a/src/replication.c
+++ b/src/replication.c
@@ -132,8 +132,8 @@ allocate_rcommon_mgmt_cmd(uint64_t buf_size)
 {
 
 	rcommon_mgmt_cmd_t *rcomm_mgmt;
-	rcomm_mgmt = (struct rcommon_mgmt_cmd *)malloc(sizeof (struct rcommon_mgmt_cmd));
-	rcomm_mgmt->buf = malloc(buf_size);
+	rcomm_mgmt = (struct rcommon_mgmt_cmd *)malloc(
+	    sizeof (struct rcommon_mgmt_cmd));
 	pthread_mutex_init(&rcomm_mgmt->mtx, NULL);
 	rcomm_mgmt->cmds_sent = 0;
 	rcomm_mgmt->cmds_succeeded = 0;
@@ -239,7 +239,6 @@ send_prepare_for_rebuild(spec_t *spec, replica_t *target_replica,
 		mgmt_cmd = malloc(sizeof (mgmt_cmd_t));
 		BUILD_REPLICA_MGMT_HDR(rmgmtio, mgmt_opcode, data_len);
 
-		replica_cnt--;
 		mgmt_cmd->io_hdr = rmgmtio;
 		mgmt_cmd->io_bytes = 0;
 		mgmt_cmd->data = (char *)malloc(data_len);
@@ -261,6 +260,7 @@ send_prepare_for_rebuild(spec_t *spec, replica_t *target_replica,
 			ret = -1;
 			rcomm_mgmt->cmds_failed++;
 		}
+		replica_cnt--;
 		goto exit;
 	}
 
@@ -271,7 +271,6 @@ send_prepare_for_rebuild(spec_t *spec, replica_t *target_replica,
 		
 		if (replica_cnt == 0)
 			break;
-		replica_cnt--;
 
 		mgmt_cmd = malloc(sizeof (mgmt_cmd_t));
 		BUILD_REPLICA_MGMT_HDR(rmgmtio, mgmt_opcode, data_len);
@@ -297,6 +296,7 @@ send_prepare_for_rebuild(spec_t *spec, replica_t *target_replica,
 			rcomm_mgmt->cmds_failed++;
 			goto exit;
 		}
+		replica_cnt--;
 	}
 
 exit:
@@ -314,9 +314,9 @@ exit:
 			spec->target_replica = NULL;
 			spec->rebuild_in_progress = false;
 		} 
-	} else {
-		assert(!replica_cnt);
-	} 
+	}
+	assert(((ret == -1) && replica_cnt) || (ret == replica_cnt));
+	
 	return ret;
 }
 
@@ -1237,7 +1237,9 @@ update_replica_status(spec_t *spec, replica_t *replica)
 		    	    " seting master_replica to NULL\n",
 		    	    replica->ip, replica->port);
 		}
-	} else if ((repl_status->state == ZVOL_STATUS_DEGRADED) &&
+	}
+#if 0
+ 	else if ((repl_status->state == ZVOL_STATUS_DEGRADED) &&
 	    (repl_status->rebuild_status == ZVOL_REBUILDING_FAILED) &&
 	    (replica == spec->target_replica)) {
 		/*
@@ -1247,7 +1249,7 @@ update_replica_status(spec_t *spec, replica_t *replica)
 		spec->rebuild_in_progress = false;
 		spec->target_replica = NULL;
 	}
-
+#endif
 	/*Trigger rebuild if possible */
 	trigger_rebuild(spec);
 	MTX_UNLOCK(&spec->rq_mtx);
--- a/src/replication.c
+++ b/src/replication.c
@@ -126,36 +126,214 @@ static int handle_mgmt_event_fd(replica_t *replica);
 }
 
 /*
- * Update_volstate enables spec and initialize io_seq number
- * if (n_replica >= c_factor)  or
- *    (n_replica >= MAX (c_factor, r_factor - c_factor + 1).
- *  n_replica = number of replica connected
- *  c_factor = consistency factor
- *  r_factor = replication_factor
+ * Case 1: Rebuild from healthy replica
+ * Case 2: Mesh rebuild i.e all replicas are downgrade
+ */
+static int 
+prepare_for_rebuild(spec_t *spec, replica_t *target_replica,
+    replica_t *healthy_replica)
+{
+	int ret = 0;
+	uint64_t num = 1;
+	replica_t *replica;
+	zvol_io_hdr_t *rmgmtio = NULL;
+	uint64_t data_len = strlen(spec->volname) + 1;
+	zvol_op_code_t mgmt_opcode = ZVOL_OPCODE_PREPARE_FOR_REBUILD;
+	mgmt_cmd_t *mgmt_cmd;
+
+	// Case 1
+	if (healthy_replica != NULL) {
+
+		mgmt_cmd = malloc(sizeof (mgmt_cmd_t));
+		BUILD_REPLICA_MGMT_HDR(rmgmtio, mgmt_opcode, data_len);
+
+		mgmt_cmd->io_hdr = rmgmtio;
+		mgmt_cmd->io_bytes = 0;
+		mgmt_cmd->data = (char *)malloc(data_len);
+		snprintf((char *)mgmt_cmd->data, data_len, "%s", spec->volname);
+		mgmt_cmd->mgmt_cmd_state = WRITE_IO_SEND_HDR;
+
+		spec->rebuild_replica_count++;  
+	
+		MTX_LOCK(&healthy_replica->r_mtx);
+		TAILQ_INSERT_TAIL(&healthy_replica->mgmt_cmd_queue, mgmt_cmd,
+		    mgmt_cmd_next);
+		MTX_UNLOCK(&healthy_replica->r_mtx);
+
+		if (write(healthy_replica->mgmt_eventfd1, &num,
+		    sizeof (num)) != sizeof (num)) {
+			REPLICA_NOTICELOG("Failed to inform to mgmt_eventfd "
+			    "for replica(%p)\n", replica);
+			ret = -1;
+		}
+		return ret;
+	}
+
+	// Case 2
+	TAILQ_FOREACH(replica, &spec->rq, r_next) {
+		if (replica == target_replica)
+			continue;
+
+		spec->rebuild_replica_count++;  
+	
+		mgmt_cmd = malloc(sizeof (mgmt_cmd_t));
+		BUILD_REPLICA_MGMT_HDR(rmgmtio, mgmt_opcode, data_len);
+
+		mgmt_cmd->io_hdr = rmgmtio;
+		mgmt_cmd->io_bytes = 0;
+		mgmt_cmd->data = (char *)malloc(data_len);
+		snprintf((char *)mgmt_cmd->data, data_len, "%s", spec->volname);
+		mgmt_cmd->mgmt_cmd_state = WRITE_IO_SEND_HDR;
+
+		MTX_LOCK(&replica->r_mtx);
+		TAILQ_INSERT_TAIL(&replica->mgmt_cmd_queue, mgmt_cmd,
+		    mgmt_cmd_next);
+		MTX_UNLOCK(&replica->r_mtx);
+		if (write(replica->mgmt_eventfd1, &num, sizeof (num)) !=
+		    sizeof (num)) {
+			REPLICA_NOTICELOG("Failed to inform to mgmt_eventfd "
+			    "for replica(%p)\n", replica);
+			return (ret = -1);
+		}
+	}
+	return ret;
+}
+
+static int
+start_rebuild(void *buf, replica_t *replica, uint64_t data_len)
+{
+	
+	int ret = 0;
+	uint64_t num = 1;
+	zvol_io_hdr_t *rmgmtio = NULL;
+	zvol_op_code_t mgmt_opcode = ZVOL_OPCODE_START_REBUILD;
+	mgmt_cmd_t *mgmt_cmd;
+
+	mgmt_cmd = malloc(sizeof (mgmt_cmd_t));
+	BUILD_REPLICA_MGMT_HDR(rmgmtio, mgmt_opcode, data_len);
+
+	mgmt_cmd->io_hdr = rmgmtio;
+	mgmt_cmd->io_bytes = 0;
+	mgmt_cmd->data = buf;
+	mgmt_cmd->mgmt_cmd_state = WRITE_IO_SEND_HDR;
+
+	MTX_LOCK(&replica->r_mtx);
+	TAILQ_INSERT_TAIL(&replica->mgmt_cmd_queue, mgmt_cmd, mgmt_cmd_next);
+	MTX_UNLOCK(&replica->r_mtx);
+
+	if (write(replica->mgmt_eventfd1, &num, sizeof (num)) !=
+	    sizeof (num)) {
+		REPLICA_NOTICELOG("Failed to inform to mgmt_eventfd "
+		    "for replica(%p)\n", replica);
+		return (ret = -1);
+	}
+	REPLICA_LOG("Start_rebuild opcode sent for Replica ip:%s port:%d "
+	    "state:%d\n", replica->ip, replica->port, replica->state);
+	return ret;
+}
+
+/* Rebuild can be of two type:-
+ * 1. Rebuilding a replica from healthy replica
+ * 2. Rebuilding a replica from all other replica(Mesh rebuild)
+ * API does the following tasks:-
+ * 1. find if there is healthy replica available for rebuilding
+ * 2. find downgraded replica for rebuild
+ * 3. Call prepare_for_rebuild to on healthy replica or on all
+ * other replicas to get IP address and port for replicas.
  */
 void
 update_volstate(spec_t *spec)
 {
+	int ret = 0;
 	uint64_t max;
-	replica_t *replica;
+	replica_t *replica = NULL;
+	replica_t *target_replica = NULL;
+	replica_t *healthy_replica = NULL;
 
-	if(((spec->healthy_rcount + spec->degraded_rcount >= spec->consistency_factor) &&
-		(spec->healthy_rcount >= 1))||
-		(spec->healthy_rcount  + spec->degraded_rcount 
-			>= MAX(spec->replication_factor - spec->consistency_factor + 1, spec->consistency_factor))) {
-		if (spec->ready == false)
-		{
-			max = 0;
-			TAILQ_FOREACH(replica, &spec->rq, r_next)
-				max = (max < replica->initial_checkpointed_io_seq) ?
-				    replica->initial_checkpointed_io_seq : max;
+	if (spec->rebuild_in_progress == true) {
+		REPLICA_NOTICELOG("Rebuild is already in progress "
+		    "on replica:%s\n", spec->volname);
+		return;
+	}
 
+	if (((spec->healthy_rcount + spec->degraded_rcount >=
+	    spec->consistency_factor) && (spec->healthy_rcount >= 1)) ||
+	    (spec->healthy_rcount  + spec->degraded_rcount >=
+	    MAX(spec->replication_factor - spec->consistency_factor + 1,
+	    spec->consistency_factor))) { 
+
+		max = 0;
+		TAILQ_FOREACH(replica, &spec->rq, r_next) {
+			/* Find healthy replica */
+			if ((healthy_replica == NULL) &&
+			    (replica->state == ZVOL_STATUS_HEALTHY)) {
+				healthy_replica = replica;
+			}
+
+			/*
+			 * If we know master replica and healthy
+			 * replica then trigger rebuild
+			 */
+			if ((spec->master_replica != NULL) &&
+			    (healthy_replica != NULL))
+				break;
+			/*
+			 * Should not account io_seq from healthy replica
+			 * when trying to find out degraded replica
+			 */
+			if (replica->state == ZVOL_STATUS_HEALTHY)
+				continue;
+			if (max < replica->initial_checkpointed_io_seq) {
+				max = replica->initial_checkpointed_io_seq;
+				target_replica = replica;
+			}
+		}
+
+		/*
+		 * Increase io_seq only when we do not have healthy replica
+		 */	
+		if (healthy_replica == NULL) {
 			max = (max == 0) ? 10 : max + (1<<20);
 			spec->io_seq = max;
 		}
 		spec->ready = true;
 	} else {
 		spec->ready = false;
+	}
+
+	REPLICA_LOG("Healthy count:%d degraded count:%d consistency factor:%d"
+	    " replication factor:%d\n", spec->healthy_rcount,
+	    spec->degraded_rcount, spec->consistency_factor,
+	    spec->replication_factor);
+
+	if (target_replica != NULL) {
+		spec->rebuild_replica_count = 0;
+		spec->rebuild_info_ack_recvd = 0;
+		spec->master_replica = target_replica;
+		spec->rebuild_in_progress = true;
+		if (spec->rebuild_info != NULL) {
+			REPLICA_ERRLOG("Freeing rebuild_info:%p\n",
+			    spec->rebuild_info);
+			free(spec->rebuild_info);
+			spec->rebuild_info = NULL;		
+		}
+		ret = prepare_for_rebuild(spec, target_replica,
+		    healthy_replica);
+		if (ret == 0) {
+			REPLICA_LOG("%s rebuild will be attempted on replica "
+			    "ip:%s port:%d state:%d\n",
+			    (healthy_replica ? "Normal" : "Mesh"),
+			    target_replica->ip, target_replica->port,
+			    target_replica->state);
+		} else {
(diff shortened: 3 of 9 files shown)


src/replication.c
```diff
@@ -126,37 +126,215 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+	replica_t *healthy_replica = NULL;
+
+	if (spec->rebuild_in_progress == true) {
+		REPLICA_NOTICELOG("Rebuild is already in progress "
+		    "on replica:%s\n", spec->volname);
+		return;
+	}
+
+	if (((spec->healthy_rcount + spec->degraded_rcount >=
+	    spec->consistency_factor) && (spec->healthy_rcount >= 1)) ||
```
vishnuitta: not related to this PR.. but, we have issue in checking for healthy_count as 1.. here, health_rcount should be checked against consistency_factor.. so, it becomes:
- healthy_rcount >= CF (or)
- (healthy_rcount + degraded_rcount >= MAX((RF-CF+1), CF)

src/replication.c
```diff
@@ -126,37 +126,215 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
 update_volstate(spec_t *spec)
 {
+	int ret = 0;
 	uint64_t max;
-	replica_t *replica;
+	replica_t *replica = NULL;
+	replica_t *target_replica = NULL;
+	replica_t *healthy_replica = NULL;
+
+	if (spec->rebuild_in_progress == true) {
```
vishnuitta, line 378: returning here will lead to issue in not updating spec's ready state. It would be better to move rebuilding functionality to a different function, and let update_volstate to update spec's ready state.

src/replication.c
```diff
@@ -126,37 +126,215 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+				target_replica = replica;
+			}
+		}
 
+		/*
+		 * Increase io_seq only when we do not have healthy replica
+		 */	
+		if (healthy_replica == NULL) {
 			max = (max == 0) ? 10 : max + (1<<20);
 			spec->io_seq = max;
```
vishnuitta: may not be related to this PR, spec's io_seq need to modified only when istgt is restarted and spec is turning 'ready' for first time. Any successive state change of spec doesn't need to update io_seq.

vishnuitta: And, if any healthy replica is there, just taking its check_pointed_io_seq and adding (1<<20) to it should be good enough.

src/replication.c
```diff
@@ -126,37 +126,215 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+			if ((healthy_replica == NULL) &&
+			    (replica->state == ZVOL_STATUS_HEALTHY)) {
+				healthy_replica = replica;
+			}
+
+			/*
+			 * If we know master replica and healthy
+			 * replica then trigger rebuild
+			 */
+			if ((spec->master_replica != NULL) &&
```
vishnuitta: we may get 'break' here even with target_replica being NULL, and, with a replica in degraded state. This is not the intended behavior of this fn if I understood correctly.

src/replication.c
```diff
@@ -126,37 +126,215 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
 	} else {
 		spec->ready = false;
 	}
+
+	REPLICA_LOG("Healthy count:%d degraded count:%d consistency factor:%d"
+	    " replication factor:%d\n", spec->healthy_rcount,
+	    spec->degraded_rcount, spec->consistency_factor,
+	    spec->replication_factor);
+
+	if (target_replica != NULL) {
```
vishnuitta: rebuilding process should start only when spec is 'ready' to take IOs (basically the condition to make spec as 'ready' should be met) or atleast one healthy replica exists.

src/replication.c
```diff
@@ -126,37 +126,215 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+
+		MTX_LOCK(&replica->r_mtx);
+		TAILQ_INSERT_TAIL(&replica->mgmt_cmd_queue, mgmt_cmd,
+		    mgmt_cmd_next);
+		MTX_UNLOCK(&replica->r_mtx);
+		if (write(replica->mgmt_eventfd1, &num, sizeof (num)) !=
+		    sizeof (num)) {
+			REPLICA_NOTICELOG("Failed to inform to mgmt_eventfd "
+			    "for replica(%p)\n", replica);
+			return (ret = -1);
```
vishnuitta: in all degraded mode case, mesh rebuild should happen among MAX(RF-CF+1, CF) number of replicas including the one on which rebuild is happening. So, 'prepare_for_rebuild' should be sent to atleast MAX(RF-CF+1, CF)-1 number of degraded replicas.
So, we need to make sure above condition meets. Also, errors from few replicas for 'prepare_for_rebuild' can be absorbed if above condition meets.
For any of these, consider the example: RF as 7, and CF as 3.

src/replication.c
```diff
@@ -487,10 +665,19 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
...
 	spec->degraded_rcount++;
 	TAILQ_REMOVE(&spec->rwaitq, replica, r_waitnext);
 	TAILQ_INSERT_TAIL(&spec->rq, replica, r_next);
+
+	MTX_UNLOCK(&spec->rq_mtx);
+
+	/*
+	 * Sleep for timeout sec so that outstanding IOs got committed to disk
+	 */
+	sleep(spec->replica_io_timeout);
```
vishnuitta: thread should not be made to 'sleep' explicitly. All we need to make sure is: rebuild should not be started for this amount of time and IOs should be started on this replica.

src/istgt_lu.h
```diff
@@ -820,12 +820,18 @@ typedef struct istgt_lu_disk_t {
...
+	int replica_io_timeout;
 	int replication_factor;
 	int consistency_factor;
 	int healthy_rcount;
 	int degraded_rcount;
 	bool ready;
+	bool rebuild_in_progress;
+	int rebuild_replica_count;
+	int rebuild_info_ack_recvd;
+	struct replica_s *master_replica;
```
vishnuitta: what is master_replica means?

vishnuitta: is this the replica that is under rebuild?

src/data_conn.c
```diff
@@ -226,6 +226,14 @@ handle_data_conn_error(replica_t *r)
 	else if (r->state == ZVOL_STATUS_DEGRADED)
 		spec->degraded_rcount--;
 
+	if (spec->master_replica == r) {
```
vishnuitta: can this be moved to the end of handle_mgmt_conn_error?

src/replication.c
```diff
@@ -831,32 +1024,78 @@ ask_replica_status_all(spec_t *spec)
...
+			spec->master_replica = NULL;
+			spec->rebuild_in_progress = false;
+			REPLICA_ERRLOG("Replica:%s port:%d marked healthy,"
+		    	    " seting master_replica to NULL\n",
+		    	    replica->ip, replica->port);
 		}
+		/* Trigger another rebuild if required */
+		update_volstate(spec);
+	} else if ((repl_status->state == ZVOL_STATUS_DEGRADED) &&
+	    (repl_status->rebuild_status == ZVOL_REBUILDING_FAILED) &&
```
vishnuitta: let us take rebuild failure and re-triggering the rebuild in next PR.

src/replication.c
```diff
@@ -831,32 +1024,78 @@ ask_replica_status_all(spec_t *spec)
...
- * This function processes the status response received from replica
- * and update spec's healthy/degrade count
- */
+static void
+update_rebuild_info(spec_t *spec, mgmt_ack_t *ack_data)
+{
+
+	uint64_t data_len = 0;
+
+	MTX_LOCK(&spec->rq_mtx);
```
vishnuitta: all the functions related to rebuild in this PR seems to be done by mgmt_thread itself. So, this lock may not be required.

src/replication.c
```diff
@@ -126,37 +126,215 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
 		spec->ready = false;
 	}
+
+	REPLICA_LOG("Healthy count:%d degraded count:%d consistency factor:%d"
+	    " replication factor:%d\n", spec->healthy_rcount,
+	    spec->degraded_rcount, spec->consistency_factor,
+	    spec->replication_factor);
+
+	if (target_replica != NULL) {
+		spec->rebuild_replica_count = 0;
```
vishnuitta: let us initialize wherever we need these, i.e, in prepare_for_rebuild()..

src/replication.c
```diff
@@ -1135,11 +1374,23 @@ read_io_resp(spec_t *spec, replica_t *replica, io_event_t *revent, mgmt_cmd_t *m
 					free(*resp_data);
 					break;
 
+				case ZVOL_OPCODE_PREPARE_FOR_REBUILD:
```
vishnuitta: Status of zvol_io_hdr (i.e., ZVOL_OP_STATUS_OK or ZVOL_OP_STATUS_FAILED etc) need to be taken into consideration.
Some other commands also need to handle this.

vishnuitta: We also need to make sure that len field in zvol_io_hdr should be 0 in case of status not being OK

src/replication.c
```diff
@@ -831,32 +1024,78 @@ ask_replica_status_all(spec_t *spec)
...
+
+	if (spec->rebuild_info == NULL)
+		spec->rebuild_info = (mgmt_ack_t *)malloc(sizeof (mgmt_ack_t) *
+		    spec->rebuild_replica_count);
+
+	memcpy(&spec->rebuild_info[spec->rebuild_info_ack_recvd], ack_data,
+	    sizeof (mgmt_ack_t));
+	spec->rebuild_info_ack_recvd++;
+
+	if (spec->rebuild_info_ack_recvd == spec->rebuild_replica_count) {
```
vishnuitta: status response which are NOT OK, and, also, connection breaks after sending 'prepare_for_rebuild' command need to be handled.
Connection break kind of errors need to be handled in empty_mgmt_q_of_replica()

src/replication.c
```diff
@@ -831,32 +1024,78 @@ ask_replica_status_all(spec_t *spec)
...
+		data_len = spec->rebuild_info_ack_recvd * sizeof (mgmt_ack_t);
+		start_rebuild(spec->rebuild_info, spec->master_replica, data_len);
```
vishnuitta: master_replica, i.e., replica that need to be rebuilt, being gone need to be handled.

vishnuitta: also, error from start_rebuild need to be handled.

src/data_conn.c
satbirchhikara: Would like to hear adavantage/disadvantage of moving/keeping it there/here before we take a decision. Please let us know!!!

src/istgt_lu.h
satbirchhikara: Yes. Will rename it.

satbirchhikara: Yes, this is replica under rebuild, renamed it to target_replica.

src/replication.c
satbirchhikara: Good catch! I also had doubts ..like do we allow replication_factor to be changed ? Was worried about if replication factor being allowed to change and say we have few NEW replicas added just a minute before iSCSI controller crashed. In that case, once iSCSI controller boot back, we may endup triggering rebuild from these Newly added replicas. Before I make change in code, lets discuss more on it in person.

src/replication.c
satbirchhikara: Sir, trigerring rebuild as well as marking it "ready=true" both are based on same condition/check. We trigger rebuild only when above condition passed, same way we marked "ready=true" if above condition passed, so IMHO these both should go into same function. 

Related to rebuild_in_progress, I would like to added assert that if rebuild_in_progress == true then this condition "ready==true" should be true. Because if rebuild_in_progress is in progress then there is all reasons that ready should be true at that time.

src/replication.c
satbirchhikara: Lets discuss it in person for better understanding/clarity.

src/replication.c
satbirchhikara: Good catch sir!!! Done!!

src/replication.c
satbirchhikara: "spec's io_seq need to modified only when istgt is restarted" this is equal to Mesh rebuild  where there is no healthy_replica in the system. In current implementation we increasing spec_io_seq only when there is no healthy_replica in the system. 

>>>And, if any healthy replica is there, just taking its check_pointed_io_seq and adding (1<<20) to it should be good enough
Sir, I did not understand this comment, why we need to increase io_seq number when there is already a healthy replica in the system.

src/replication.c
satbirchhikara: Sir target_replica will be non-null if we have gone through consistency factor check and ideally we should have marked spec->ready=true. Will add assert under "if (target_replica != NULL)" to make sure that if we are here it means spec is ready for IOs.

src/replication.c
satbirchhikara: Done sir!!!

src/replication.c
satbirchhikara: Yes, that is the requirement. We can offload triggering of rebuild on newly added/registered replica  to ask_replica_status_all function, with a time stamp that it should trigger update_volstate on this replica not before TIME_OUT elapsed. In this case we have to added time_stamp in replica the time it got registered with iSCSI controller. Any suggestion/thought ?

src/replication.c
satbirchhikara: Sir, lock/unlock can be avoided for perfromance reasons. There is saying that if you confused, go for lock :-). As you can see that these functions will be called once in a while i.e. locks are not
in that perfromance critical path. I urge you to reconsider this comment.

Other thing is that first going without lock and then revisiting code to implement lock is very difficult task compare to reverse i.e. going with lock in first attempt and then revisiting.

src/replication.c
satbirchhikara: Status NOT_OK need to be handled, I agree, but connection break be handled automatically when we call handle_mgmt_conn_error/handle_data_conn_error which I think bound to be called when there is problem w.r.t either connection.

src/replication.c
satbirchhikara: Sir, replica being rebuilt and gone handled in handle_data_conn_error. 

if (spec->target_replica == r) {
}

With respect to handling error from start_rebuild, not sure what can we do. Do we want to go back to pick another replica ? If error came on fd, there should be a reason which might be caught through other code paths like dis-connect etc. No ?

src/replication.c
satbirchhikara: Next PR, you mean remove this code from this PR ?

src/replication.c
satbirchhikara: Yes sir. Will try to make it uniform for all cases, handle_snap_create_resp is little different which need to be discussed.

satbirchhikara (commented):
Vitta Sir,

I tried my best to resolve comments. I may have missed providing/explaining. Feel free to comment back. I need a discussion to resolve few of the comments(where I am not clear or there is ambiguity).

Thanks & Regards,
Satbir Singh

src/replication.c
vishnuitta: sure..
basically, idea is: conditions need to be developed in such a way that data should be able to rebuild finally with the given RF and CF.

src/replication.c
vishnuitta: ok.. lets change this to assert

src/replication.c
vishnuitta: regarding to setting io_seq (as algorithm), scenarios I'm thinking are of:
case 1: istgt is restarted, and, volume is becoming 'not ready' to 'ready state, after required number of replicas are connected in degraded state.
case 2: istgt is running. It went from 'ready' state to 'not ready' state due to replicas disconnection. It again became 'ready', but, there are no healthy replicas.
case 3: istgt is running. It went from 'ready' state to 'not ready' state due to replicas disconnection. It again became 'ready', with atleast a healthy replica in it.
For second and third cases, we don't need to change io_seq, as istgt already knows the io_seq, and it don't wrote any other IOs.
For first case, looking out for max, and adding (1<<20) should be fine.
As you mentioned, healthy replica case is not needed to handle. I mentioned that comment wrt case 3, and, check_pointed_io_seq being old than the latest one.

vishnuitta: regarding to this PR, we are not looking for transition from 'not ready' to 'ready'. This can lead to regular updates of io_seq. Is check for 'spec->ready' being false removed intentionally?
Combining spec's 'ready' with rebuilding seems to be making this logic complex. If things can go hand in hand, it should be fine leaving them together.

src/replication.c
vishnuitta: Another issue with 'break' can be is: we may not be finding the 'max' to set io_seq.

vishnuitta: ok.. regarding my above comment on issue in finding 'max' to set io_seq, I got your point that - with healthy replica being there, we don't need to update io_seq. We will be in 'case 3' of scenarios explained in other comment.

vishnuitta: is master_replica supposed to be NULL here as 'rebuild_in_progress' is false?

src/replication.c
vishnuitta: yes.. timestamp that need to check for next rebuild is good to have. This can also make code not to piggyback on ask_replica_status_all function, but, can have its own if needed.

src/data_conn.c
vishnuitta: mgmt_conn is the one on which rebuild commands are sent, and, handle_mgmt_conn_error tears down the replica completely. So, it would be good to move to that fn.

src/replication.c
```diff
@@ -126,37 +126,215 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+		}
+		return ret;
+	}
+
+	// Case 2
+	TAILQ_FOREACH(replica, &spec->rq, r_next) {
+		if (replica == target_replica)
+			continue;
+
+		spec->rebuild_replica_count++;
```
vishnuitta: similar to rcmd_t (which is about data command sent to replica) and rcomm_cmd_t (which is common structure contains details about all rcmd_t's sent to replicas), mgmt_cmd (mgmt command sent to replica) have rcommon_mgmt_cmd_t. This can be helpful here, to replace newly added rebuild related variables of spec.

src/replication.c
vishnuitta: for MVP also, there is requirement to add new replica, which means, replication factor need to be increased. mgmt and data paths are covered for this requirement. So, it would be good to cover.
But, there are still pending things to do with respect to adding/updating replica to the list. Thanks for pointing to this. Created a issue to track this under release.

src/replication.c
vishnuitta: oh ok.. got it.. you are right sir.
as you said, assert would be needed for any further modifications between these two chunks of code.

vishnuitta: or for better clarity (which currently gives impression that target_replica should be non-null for rebuild), how about moving this under the case where spec's ready is made as true?

src/replication.c
vishnuitta: I got your point sir..

src/replication.c
vishnuitta: yes.. handle_mgmt_conn_error/handle_data_conn_err are called during connection break, and those clears the mgmt_q pending commands. In those functions, handlers need to be called for new mgmt commands (in empty_mgmt_q_of_replica fn)

src/replication.c
vishnuitta: spec's target_replica has been set to NULL when it disconnects, but, it need to be checked here.

vishnuitta: error from start_rebuild should restart the entire process of rebuild again.
disconnect path takes care of removing replica and its any pending IOs. But, here, error came on target_replica on which there are no pending mgmt_cmds.
Things like rebuild_in_progress etc need to be set to false.

src/replication.c
vishnuitta: There are few pending things in re-triggering rebuild like detecting that rebuild threads are got destroyed at replica. 
And, to give focus on triggering rebuild in this PR, I suggest to move re-triggering rebuild part to next PR.

src/replication.c
vishnuitta: ok sir..

src/data_conn.c
satbirchhikara: Done.

src/replication.c
satbirchhikara: Done!

src/replication.c
satbirchhikara: I have done the changes, Please have a look sir!!

src/replication.c
satbirchhikara: Lots of code changes happened. Lets see if comment is valid w.r.t current changes,

satbirchhikara (commented):
Hi Vitta,

I have done change as per comments received. Please have a look and let me know your comments.

Thanks & regards,
Satbir Singh

src/replication.c
```diff
@@ -125,37 +127,330 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+	uint64_t max = 0;
+	struct timespec now, diff;
+	replica_t *replica = NULL;
+	replica_t *target_replica = NULL;
+	replica_t *healthy_replica = NULL;
+
+	if (spec->rebuild_in_progress == true) {
+		REPLICA_NOTICELOG("Rebuild is already in progress "
+		    "on volume:%s\n", spec->volname);
+		MTX_UNLOCK(&spec->rq_mtx);
```
vishnuitta: we are unlocking in the caller.

src/replication.c
```diff
@@ -125,37 +127,330 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
-				    replica->initial_checkpointed_io_seq : max;
+		if (max < replica->initial_checkpointed_io_seq) {
+			max = replica->initial_checkpointed_io_seq;
+			target_replica = replica;
+		}
 
+		/*
+		 * Increase io_seq only when we do not have healthy replica
+		 */	
+		if (healthy_replica == NULL) {
```
vishnuitta: we don't need this here now..

vishnuitta: basically, updating io_seq of spec..

src/replication.c
```diff
@@ -125,37 +127,330 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
 		}
+	}
+
+	REPLICA_LOG("Healthy count:%d degraded count:%d consistency factor:%d"
+	    " replication factor:%d\n", spec->healthy_rcount,
+	    spec->degraded_rcount, spec->consistency_factor,
+	    spec->replication_factor);
+
+	if (target_replica == NULL) {
+		assert(!"Can this happen ?\n");
```
vishnuitta: this can happen when all replicas got just connected

src/replication.c
```diff
@@ -125,37 +127,330 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
-		(spec->healthy_rcount >= 1))||
-		(spec->healthy_rcount  + spec->degraded_rcount 
-			>= MAX(spec->replication_factor - spec->consistency_factor + 1, spec->consistency_factor))) {
-		if (spec->ready == false)
-		{
-			max = 0;
-			TAILQ_FOREACH(replica, &spec->rq, r_next)
-				max = (max < replica->initial_checkpointed_io_seq) ?
-				    replica->initial_checkpointed_io_seq : max;
+		if (max < replica->initial_checkpointed_io_seq) {
```
vishnuitta: lets make the check as '<='..

src/replication.c
```diff
@@ -125,37 +127,330 @@ static int handle_mgmt_event_fd(replica_t *replica);
 	}								\
 }
 
+static rcommon_mgmt_cmd_t *
+allocate_rcommon_mgmt_cmd(uint64_t buf_size)
+{
+
+	rcommon_mgmt_cmd_t *rcomm_mgmt;
+	rcomm_mgmt = (struct rcommon_mgmt_cmd *)malloc(sizeof (struct rcommon_mgmt_cmd));
+	rcomm_mgmt->buf = malloc(buf_size);
```
vishnuitta: this is not required..

src/replication.c
```diff
@@ -125,37 +127,330 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+	spec->target_replica = target_replica;
+	spec->rebuild_in_progress = true;
+
+	replica_cnt = count_of_replicas_helping_rebuild(spec, healthy_replica);
+	size = replica_cnt * sizeof (mgmt_ack_t); 
+	rcomm_mgmt = allocate_rcommon_mgmt_cmd(size);
+	// Case 1
+	if (healthy_replica != NULL) {
+
+		mgmt_cmd = malloc(sizeof (mgmt_cmd_t));
```
vishnuitta: lets set replica to health_replica so that 'replica' can be used in exit code

src/replication.c
```diff
@@ -125,37 +127,330 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+	}
+
+exit:
+	if (ret == -1) {
+		MTX_LOCK(&replica->r_mtx);
+		TAILQ_REMOVE(&replica->mgmt_cmd_queue, mgmt_cmd,
+		    mgmt_cmd_next);
+		MTX_UNLOCK(&replica->r_mtx);
+		if (rcomm_mgmt->cmds_failed == rcomm_mgmt->cmds_sent) {
+			free(mgmt_cmd);
```
vishnuitta: mgmt_cmd can be freed outside this 'if' condition

src/replication.c
```diff
@@ -125,37 +127,330 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+		if (write(replica->mgmt_eventfd1, &num, sizeof (num)) !=
+		    sizeof (num)) {
+			REPLICA_NOTICELOG("Failed to inform to mgmt_eventfd "
+			    "for replica(%p)\n", replica);
+			ret = -1;
+			rcomm_mgmt->cmds_failed++;
+			goto exit;
+		}
+	}
+
```
vishnuitta: lets have an assert that 'replica_count' should be 0 here

src/replication.c
```diff
@@ -125,37 +127,330 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+exit:
+	if (ret == -1) {
+		MTX_LOCK(&replica->r_mtx);
```
vishnuitta: lets add the reason why we can remove from queue here, so that, any references in future for this will not copy this as it is..

src/replication.c
```diff
@@ -125,37 +127,330 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+	// Case 2
+	TAILQ_FOREACH(replica, &spec->rq, r_next) {
+		if (replica == target_replica)
+			continue;
+		
+		if (replica_cnt == 0)
+			break;
+		replica_cnt--;
+
+		mgmt_cmd = malloc(sizeof (mgmt_cmd_t));
```
vishnuitta: how about making this code as a fn or macro?

src/replication.c
```diff
@@ -528,8 +825,12 @@ send_replica_handshake_query(replica_t *replica, spec_t *spec)
 	MTX_LOCK(&replica->r_mtx);
 	TAILQ_INSERT_TAIL(&replica->mgmt_cmd_queue, mgmt_cmd, mgmt_cmd_next);
 	MTX_UNLOCK(&replica->r_mtx);
+	if (write(replica->mgmt_eventfd1, &num, sizeof (num)) != sizeof (num)) {
+		REPLICA_NOTICELOG("Failed to inform to mgmt_eventfd for replica(%p)\n", replica);
```
vishnuitta: failed one need to be removed from linked list

vishnuitta: or let us leave it as it is.. and instead return 0

src/replication.c
```diff
@@ -1250,11 +1613,15 @@ empty_mgmt_q_of_replica(replica_t *r)
 {
 	mgmt_cmd_t *mgmt_cmd;
 	while ((mgmt_cmd = TAILQ_FIRST(&r->mgmt_cmd_queue))) {
+		mgmt_cmd->io_hdr->status = ZVOL_OP_STATUS_FAILED;
 		switch (mgmt_cmd->io_hdr->opcode) {
 			case ZVOL_OPCODE_SNAP_CREATE:
-				mgmt_cmd->io_hdr->status = ZVOL_OP_STATUS_FAILED;
 				handle_snap_create_resp(r, mgmt_cmd);
```
vishnuitta, line 1659: not related to this PR..
As you found through code walk-through, lets add a comment that handle_snap_create_resp is picking header status from different header

src/replication.c
```diff
@@ -125,37 +127,330 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+		}
+	}
+
+exit:
+	if (ret == -1) {
+		MTX_LOCK(&replica->r_mtx);
+		TAILQ_REMOVE(&replica->mgmt_cmd_queue, mgmt_cmd,
+		    mgmt_cmd_next);
+		MTX_UNLOCK(&replica->r_mtx);
+		if (rcomm_mgmt->cmds_failed == rcomm_mgmt->cmds_sent) {
```
vishnuitta: lets also add comment that why we don't need lock of rcomm_mgmt to access these variables

src/replication.c
```diff
@@ -831,34 +1127,93 @@ ask_replica_status_all(spec_t *spec)
...
+			spec->rebuild_in_progress = false;
+			MTX_UNLOCK(&spec->rq_mtx);
+			REPLICA_LOG("Unable to trigger rebuild on Replica ip:"
+			    "%s port:%d state:%d\n", spec->target_replica->ip,
+			    spec->target_replica->port, spec->target_replica->state);
+		}	
+		free_rcommon_mgmt_cmd(rcomm_mgmt);
+	} else if (rcomm_mgmt->cmds_sent ==
+	    (rcomm_mgmt->cmds_failed + rcomm_mgmt->cmds_succeeded)) {
+		free_rcommon_mgmt_cmd(rcomm_mgmt);
```
vishnuitta, line 1195: some print over here, and, above 'NOT_OK' case

src/replication.c
```diff
@@ -831,34 +1127,93 @@ ask_replica_status_all(spec_t *spec)
...
+
+	if (rcomm_mgmt->cmds_sent == rcomm_mgmt->cmds_succeeded) {
+		ret = start_rebuild(buf, spec->target_replica, rcomm_mgmt->buf_size);
+		if (ret == 0) {
+			rcomm_mgmt->buf = NULL;
+			REPLICA_LOG("Rebuild triggered on Replica ip:%s port:%d"
+			    " state:%d\n", spec->target_replica->ip,
+			    spec->target_replica->port, spec->target_replica->state);
+		} else {
+			MTX_LOCK(&spec->rq_mtx);
```
vishnuitta, line 1184: we won't be needing this lock..

src/replication.c
```diff
@@ -125,37 +127,330 @@ static int handle_mgmt_event_fd(replica_t *replica);
...
+	mgmt_cmd->data = buf;
+	mgmt_cmd->mgmt_cmd_state = WRITE_IO_SEND_HDR;
+
+	MTX_LOCK(&replica->r_mtx);
+	TAILQ_INSERT_TAIL(&replica->mgmt_cmd_queue, mgmt_cmd, mgmt_cmd_next);
+	MTX_UNLOCK(&replica->r_mtx);
+
+	if (write(replica->mgmt_eventfd1, &num, sizeof (num)) !=
+	    sizeof (num)) {
+		REPLICA_NOTICELOG("Failed to inform to mgmt_eventfd "
```
vishnuitta, line 347: lets remove from linked list

src/replication.c
```diff
@@ -831,34 +1127,93 @@ ask_replica_status_all(spec_t *spec)
...
+			rcomm_mgmt->buf = NULL;
+			REPLICA_LOG("Rebuild triggered on Replica ip:%s port:%d"
+			    " state:%d\n", spec->target_replica->ip,
+			    spec->target_replica->port, spec->target_replica->state);
+		} else {
+			MTX_LOCK(&spec->rq_mtx);
+			spec->target_replica = NULL;
+			spec->rebuild_in_progress = false;
+			MTX_UNLOCK(&spec->rq_mtx);
+			REPLICA_LOG("Unable to trigger rebuild on Replica ip:"
```
vishnuitta, line 1188: need to handle cleanup of mgmt_cmd

src/replication.c
vishnuitta: there is a fn clear_mgmt_cmd to free mgmt_cmd.. please check whether that can be used here

src/replication.c
```diff
@@ -831,34 +1127,93 @@ ask_replica_status_all(spec_t *spec)
...
+			/* master_replica became healthy*/
 			spec->degraded_rcount--;
 			spec->healthy_rcount++;
+			spec->target_replica = NULL;
+			spec->rebuild_in_progress = false;
+			REPLICA_ERRLOG("Replica:%s port:%d marked healthy,"
+		    	    " seting master_replica to NULL\n",
+		    	    replica->ip, replica->port);
 		}
-		update_volstate(spec);
```
vishnuitta, line 860: we would be needing this

src/replication.c
```diff
@@ -1109,37 +1467,42 @@ read_io_resp(spec_t *spec, replica_t *replica, io_event_t *revent, mgmt_cmd_t *m
...
 					break;
 
 				case ZVOL_OPCODE_SNAP_CREATE:
+			
+					assert(fd != replica->iofd);
 					/* snap create response must come from mgmt connection */
 					handle_snap_create_resp(replica, mgmt_cmd);
 					break;
 
+				case ZVOL_OPCODE_START_REBUILD:
```
vishnuitta, line 1541: handler for start_rebuild

src/replication.h
```diff
@@ -166,7 +168,7 @@ int64_t perform_read_write_on_fd(int fd, uint8_t *data, uint64_t len,
 int initialize_volume(spec_t *spec, int, int);
 
 /* Replica default timeout is 200 seconds */
-#define	REPLICA_DEFAULT_TIMEOUT	200
+#define	REPLICA_DEFAULT_TIMEOUT	1
```
vishnuitta: lets not change this.. let us make a new variable, and change it from test code

src/data_conn.c
```diff
@@ -226,6 +226,14 @@ handle_data_conn_error(replica_t *r)
 	else if (r->state == ZVOL_STATUS_DEGRADED)
 		spec->degraded_rcount--;
 
+	if (spec->target_replica == r) {
```
vishnuitta: this is better to be at handle_mgmt_conn_error..

vishnuitta (commented):
Overall changes are good. Left few comments. Please check if they are valid.
Single replica rebuild is pending.

src/data_conn.c
satbirchhikara: Done.

src/replication.c
satbirchhikara: Why sir ? We are using this buf to append <Rebuild_IP and Port> for replicas in prepare_for_rebuild_response.

src/replication.c
satbirchhikara: Done sir!!

src/replication.c
satbirchhikara: Sir, puting it in macro/function will defeat purpose unless we have more instances of same code or atleast I am not aware of(if already exist).  Anyway we have few more macros which we need to put in place/shape(to improve code readability), I would like to revist this code via separate PR for these minor enhancements.

src/replication.c
satbirchhikara: Replica_cnt may be zero and non_zero in failure case while as in success case, it should be zero. I am adding assert for success case, for failure, anyway we have encountered failure.

src/replication.c
satbirchhikara: Done sir!

src/replication.c
satbirchhikara: IMHO, we should not spil comments all over the code where rcomm_mgmt is used. This should be trivial to programmer when to use mtx and when to not. I tried to add comment in struct definition itself to make it clear that, beware of using mtx if this struct is used among threads.

satbirchhikara: Done!

src/replication.c
satbirchhikara: Removed and freed.

src/replication.c
satbirchhikara: Done.

src/replication.c
satbirchhikara: Good catch sir! I have removed it.

src/replication.c
satbirchhikara: Done sir.

src/replication.c
satbirchhikara: Done sire! Check if target_replica is null then return from here.

src/replication.c
satbirchhikara: Removed assert, returning from here.

src/replication.c
satbirchhikara: Freeing it now.

src/replication.c
satbirchhikara: All over the places, I am updating this "spec" struct under this lock.  Removing it from this place will not be consistent with coding. 

There could be a separate discussion that why I am using lock to update this struct. We can discuss that and decide to remove lock from all the place where I am updating spec.

src/replication.c
satbirchhikara: It will be freed in handle_read_data_event or in empty_mgmt_q_of_replica.

src/replication.c
satbirchhikara: Done!

src/replication.c
satbirchhikara: I think update_volstate should be used only at two places ....
> during replica registration
> during replica disconnection
These are the events which can increase/decrease degraded_counts, so needed there only. Thats where we set spec->ready to true.

src/replication.c
satbirchhikara: Done!

src/replication.c
satbirchhikara: I have added it in my TODO list. Will share it with you.

src/replication.h
satbirchhikara: Ok

satbirchhikara (commented):
Hi Sir,

Replied to your comments. Will push changes soon. Please have a look and let me know if I missed addressing any of your comment.

Thanks & Regards,
Satbir Singh

src/replication.c
vishnuitta: we are allocating at the end of this fn also (second time)

src/replication.c
vishnuitta: agreed sir on the comments part. I think - free would be still required outside 'if' right?

src/replication.c
vishnuitta: for failure case, we never come here, as it does 'goto exit'

src/replication.c
vishnuitta: revisiting also fine. May not be macro, but, a fn would be good as this fn became very big, and have repeated code.

src/replication.c
vishnuitta: hmm.. lets leave it as it is and return 0, because, epollout will come right after adding to epoll and that will take care of sending out the command even though this write is failed.

src/replication.c
vishnuitta: actually, we are not using lock to update these variables at any place other than here. We are safe to remove lock here also.

src/replication.c
vishnuitta: you are right sir.. This is a messed up comment.

src/replication.c
vishnuitta: oops.. I forgot again. you are right sir.

src/replication.c
```diff
@@ -1966,6 +2006,14 @@ handle_mgmt_conn_error(replica_t *r, int sfd, struct epoll_event *events, int ev
 	mgmtfd = r->mgmt_fd;
 	r->mgmt_fd = -1;
 	MTX_UNLOCK(&r->r_mtx);
+
+	if (r->spec->target_replica == r) {
```
vishnuitta: this need to be done after 'pthread_cond_wait' which waits for data conn thread to close, along with respond_with_error_for_all_outstanding_mgmt_ios..

src/replication.c
satbirchhikara: These are effects of long reviews :-) Done sir!!

src/replication.c
satbirchhikara: Done sir!!

src/replication.c
satbirchhikara: Done!!

src/replication.c
satbirchhikara: Made a comment above rcomm_mgmt struct that when to use mtx and when to not.

src/replication.c
satbirchhikara: Code has changed a lot sir. Can you please have a look into most recent changes and let me know if this comment still valid w.r.t current changes.

src/replication.c
satbirchhikara: Do you see any harm freeing it here ? If there is no issue with freeing it here, we should free it explicitely then and there in error case to avoid wrong precedence.

src/replication.c
satbirchhikara: Sir, I hope we are talking of spec->target_replica and spec->rebuild_in_progress variables.
We are taking lock to update these variables, either caller is taking lock or these functions taking lock explicitely before updating these variable. Few of them are listed here 

1.send_prepare_for_rebuild (caller taking lock....update_replica_status)
2. handle_start_rebuild_resp
3. trigger_rebuild (caller taking lock)
4. handle_prepare_for_rebuild_resp
and many more ...

We have to added assert in all those function where lock is taken by caller to assert that when we are updating these variables, we have lock. Once Mayank's PR(assert/lock/verify related PR) will be merged, I will add assert here. Its in my todo list.

src/replication.c
satbirchhikara: Done sir. Will comment out that code.

src/replication.c
satbirchhikara: Sir do you see any correctness issue doing it at this place. I did purposely here because here we have spec->rq_mtx lock. I did not want it to be taken again. By the time we come to pthread_cond_wait, we release this lock, it means I have to take this lock again.

vishnuitta (approved):
changes are good. Thanks @satbirchhikara

PR #39, opened 2018-06-13
TA637 : debug logs for replica interface

mynktl:
**Added debug logs for replica interface**
 from log, replica can be identified using `replica(zvol_guid)` or `replica(replica_ip:replica_port)`

### DIFF
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -58,6 +58,11 @@ int replica_timeout = REPLICA_DEFAULT_TIMEOUT;
 		rcomm_cmd->resp_list[idx].status |= RECEIVED_ERR;	\
 		if (rcomm_cmd->state != CMD_EXECUTION_DONE)		\
 			pthread_cond_signal(_cond);			\
+		REPLICA_ERRLOG("command %s(offset:%lu len:%lu seq:%lu)" \
+		    " failed for replica(%lu)\n",			\
+		    (rcomm_cmd->opcode == ZVOL_OPCODE_WRITE) ? "write" :\
+		    "read", rcomm_cmd->offset, rcomm_cmd->data_len,	\
+		    rcomm_cmd->io_seq, r->zvol_guid);			\
 		free(rcmd->iov_data);					\
 		put_to_mempool(&rcmd_mempool, rcmd);			\
 		rcmd = next_rcmd;					\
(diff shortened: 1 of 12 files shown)


src/data_conn.c
```diff
@@ -58,6 +58,11 @@ int replica_timeout = REPLICA_DEFAULT_TIMEOUT;
 		rcomm_cmd->resp_list[idx].status |= RECEIVED_ERR;	\
 		if (rcomm_cmd->state != CMD_EXECUTION_DONE)		\
 			pthread_cond_signal(_cond);			\
+		REPLICA_ERRLOG("command %s(offset:%lu len:%lu seq:%lu)" \
```
satbirchhikara: If there are thousand cmds in pending queue, it will print all of them. IMO, that is too much to ask specially when this macro is called from respond_with_error_for_all_outstanding_ios() which is called when we tear down replica. I am not sure in this particular case how these logs help us debug problem.

satbirchhikara (approved):
LGTM

src/data_conn.c
```diff
@@ -58,30 +66,43 @@ int replica_timeout = REPLICA_DEFAULT_TIMEOUT;
...
-			timesdiff(pending_cmd->queued_time, now, _diff);\
+			clock_gettime(CLOCK_MONOTONIC, &nw);		\
+			timesdiff(CLOCK_MONOTONIC,			\
+			    pending_cmd->queued_time, nw, _diff);	\
 			if (_diff.tv_sec >= replica_timeout) {		\
 				REPLICA_ERRLOG("timeout happened for "	\
-				    "replica(%s:%d)..time(%lu)\n", 	\
-				    r->ip, r->port, _diff.tv_sec); 	\
+				    "replica(%lu).. delay(%lu sec)\n", 	\
+				    r->zvol_guid, _diff.tv_sec); 	\
```
vishnuitta, line 94: ip and port also should be good

src/istgt_misc.h
```diff
@@ -199,4 +199,31 @@ void istgt_yield(void);
 void poolinit(void);
 void poolfini(void);
 int poolprint(char *inbuf, int len);
+
+#define timesdiffX(_st, _now, _re)					\
```
vishnuitta: this macro can be removed completely

src/istgt.c
```diff
@@ -2633,6 +2643,38 @@ void *timerfn(void *ptr __attribute__((__unused__)))
...
 		while((conn = (CONN *)(istgt_queue_dequeue(&backupconns))) != NULL)
 			istgt_queue_enqueue(&closedconns, conn);
+
+		clock_gettime(clockid, &now);
+		timesdiff(clockid, last_check, now, diff);
+		ms = diff.tv_sec * 1000;
+		ms += diff.tv_nsec / 1000000;
+
+		if (ms > check_interval1) {
+			check_interval1 = check_interval2;
```
vishnuitta: why do we need check_interval2? can you put some comments in the logic?
Also, put some comments in data_conn.c for logic.

src/istgt.c
```diff
@@ -2633,6 +2643,38 @@ void *timerfn(void *ptr __attribute__((__unused__)))
...
+					clock_gettime(clockid, &now);
+					timesdiff(clockid, lu_cmd->times[0], now, diff);
+					ms = diff.tv_sec * 1000;
+					ms += diff.tv_nsec / 1000000;
+					if (ms > check_interval2) {
+						ISTGT_NOTICELOG("LU:%lu CSN:0x%x TT:%x OP:%2.2x:%x:%s(%lu+%u) not responded since %d seconds\n",
+						    lu_cmd->lun, lu_cmd->CmdSN, lu_cmd->task_tag, lu_cmd->cdb0, lu_cmd->status, lu_cmd->info, lu_cmd->lba, lu_cmd->lblen, ms / 1000);
+					}
+					if (ms < check_interval1)
+						check_interval1 = ms;
```
vishnuitta: this doesn't seem to be correct. With this, there will be frequent checking if I'm not wrong.
Most of the times, queue will have IOs which are less than 1 sec. So, this check will be happening more than once in a second even though print doesn't come.

src/data_conn.c
mynktl: ok

src/istgt_misc.h
mynktl: yes. it is not being used anywhere. I will remove it.

src/istgt.c
mynktl: that's right. We don't need to update `check_interval1` here.

src/istgt.c
mynktl: ok

mynktl: `check_interval2` can be removed.

mynktl:
Hi @vishnuitta 
I have addressed review comments given by you. Please have a look and let me know if any more changes are required.

vishnuitta (approved):
Changes are good.

PR #40, opened 2018-06-14
Limit the number of replicas connected to target based on RF

mynktl:
Target will allow only `replication_factor` number of replicas.
Target will disconnect any extra replica only after handshake(on mgmt connection and data connection) happens. As of now, target only loges the additional replica information to syslog.

### DIFF
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -55,9 +55,13 @@ int replica_timeout = REPLICA_DEFAULT_TIMEOUT;
 		rcomm_cmd->resp_list[idx].io_resp_hdr.status =		\
 		    ZVOL_OP_STATUS_FAILED;				\
 		rcomm_cmd->resp_list[idx].data_ptr = NULL;		\
-		rcomm_cmd->resp_list[idx].status |= RECEIVED_ERR;	\
-		if (rcomm_cmd->state != CMD_EXECUTION_DONE)		\
+		if (rcomm_cmd->state != CMD_EXECUTION_DONE) {		\
+			rcomm_cmd->resp_list[idx].status |= 		\
+			    RECEIVED_ERR;				\
 			pthread_cond_signal(_cond);			\
+		} else							\
+			rcomm_cmd->resp_list[idx].status |= 		\
+			    RECEIVED_ERR;				\
 		free(rcmd->iov_data);					\
 		put_to_mempool(&rcmd_mempool, rcmd);			\
 		rcmd = next_rcmd;					\
@@ -200,6 +204,7 @@ handle_data_conn_error(replica_t *r)
 	int fd, data_eventfd, mgmt_eventfd2, epollfd;
 	spec_t *spec;
 	replica_t *r1 = NULL;
+	int update_spec_status = 0;
 
 	if (r->iofd == -1) {
 		REPLICA_ERRLOG("repl %s %d %p\n", r->ip, r->port, r);
@@ -210,22 +215,20 @@ handle_data_conn_error(replica_t *r)
 	spec = r->spec;
 
 	TAILQ_FOREACH(r1, &(spec->rq), r_next)
-		if (r1 == r)
+		if (r1 == r) {
+			update_spec_status= 1;
 			break;
-
-	if (r1 == NULL) {
-		REPLICA_ERRLOG("replica %s %d not part of rqlist..\n",
-		    r->ip, r->port);
-		MTX_UNLOCK(&spec->rq_mtx);
-		return -1;
-	}
-	TAILQ_REMOVE(&spec->rq, r, r_next);
+		}
 
 	if (r->state == ZVOL_STATUS_HEALTHY)
 		spec->healthy_rcount--;
 	else if (r->state == ZVOL_STATUS_DEGRADED)
 		spec->degraded_rcount--;
-	update_volstate(r->spec);
+
+	if (update_spec_status) {
+		TAILQ_REMOVE(&spec->rq, r, r_next);
+		update_volstate(r->spec);
+	}
 
 	mgmt_eventfd2 = r->mgmt_eventfd2;
 
@@ -469,11 +472,12 @@ start:
 		rcomm_cmd->resp_list[idx].data_ptr = r->ongoing_io_buf;
 		if (rcomm_cmd->opcode == ZVOL_OPCODE_WRITE)
 			r->replica_inflight_write_io_cnt -= 1;
-		rcomm_cmd->resp_list[idx].status |= RECEIVED_OK;
 
 		if (rcomm_cmd->state != CMD_EXECUTION_DONE) {
+			rcomm_cmd->resp_list[idx].status |= RECEIVED_OK;
 			pthread_cond_signal(cond_var);
-		}
+		} else
+			rcomm_cmd->resp_list[idx].status |= RECEIVED_OK;
 
 		free(r->ongoing_io->iov_data);
 		put_to_mempool(&rcmd_mempool, r->ongoing_io);
@@ -541,7 +545,7 @@ replica_thread(void *arg)
 	if (r_epollfd < 0) {
 		REPLICA_ERRLOG("epoll_create error for replica(%s:%d) "
 		    "errno(%d)\n", r->ip, r->port, errno);
-		return NULL;
+		goto initialize_error;
 	}
 
 	ev.events = EPOLLIN;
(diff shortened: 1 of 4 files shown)


src/data_conn.c
```diff
@@ -210,22 +215,20 @@ handle_data_conn_error(replica_t *r)
...
-		REPLICA_ERRLOG("replica %s %d not part of rqlist..\n",
-		    r->ip, r->port);
-		MTX_UNLOCK(&spec->rq_mtx);
-		return -1;
-	}
-	TAILQ_REMOVE(&spec->rq, r, r_next);
+		}
 
 	if (r->state == ZVOL_STATUS_HEALTHY)
 		spec->healthy_rcount--;
```
satbirchhikara: Would be good to have assert here that if replica is healthy then it should be part of spec->rq.

src/data_conn.c
```diff
@@ -210,22 +215,20 @@ handle_data_conn_error(replica_t *r)
 	spec = r->spec;
 
 	TAILQ_FOREACH(r1, &(spec->rq), r_next)
-		if (r1 == r)
+		if (r1 == r) {
+			update_spec_status= 1;
```
satbirchhikara: Minor ...add space before "="

src/data_conn.c
```diff
@@ -469,11 +472,12 @@ handle_epoll_in_event(replica_t *r)
 		rcomm_cmd->resp_list[idx].data_ptr = r->ongoing_io_buf;
 		if (rcomm_cmd->opcode == ZVOL_OPCODE_WRITE)
 			r->replica_inflight_write_io_cnt -= 1;
-		rcomm_cmd->resp_list[idx].status |= RECEIVED_OK;
 
 		if (rcomm_cmd->state != CMD_EXECUTION_DONE) {
+			rcomm_cmd->resp_list[idx].status |= RECEIVED_OK;
```
satbirchhikara: Good to have comment with explanation why we set status in this way ....explain that recomm cleaner thread might free this struct if we set it before check state.

src/replication.c
```diff
@@ -436,6 +436,23 @@ trigger_rebuild(spec_t *spec)
 	}
 }
 
+static int
+verify_replica_count(spec_t *spec, replica_t *replica)
+{
+	int replica_count;
+	int ret = 0;
+
```
satbirchhikara: Good to have assert that spec->mtx is held when we are here.

src/replication.c
```diff
@@ -1940,6 +1966,7 @@ replicate(ISTGT_LU_DISK *spec, ISTGT_LU_CMD_Ptr cmd, uint64_t offset, uint64_t n
 
 			MTX_LOCK(&spec->rq_mtx);
 			TAILQ_REMOVE(&spec->rcommon_waitq, rcomm_cmd, wait_cmd_next);
+			put_to_mempool(&spec->rcommon_deadlist, rcomm_cmd);
```
satbirchhikara: Since mempool is lockless multithread safe data structure, please move  put_to_mempool call after releasing lock.

satbirchhikara (commented):
Hi Mayank,

Reveiwed it, gave few minor comments. Please have a look and let me know if you need further clarification.

Thanks & Regards,
Satbir Singh

src/data_conn.c
```diff
@@ -55,9 +55,13 @@ int replica_timeout = REPLICA_DEFAULT_TIMEOUT;
 		rcomm_cmd->resp_list[idx].io_resp_hdr.status =		\
 		    ZVOL_OP_STATUS_FAILED;				\
 		rcomm_cmd->resp_list[idx].data_ptr = NULL;		\
-		rcomm_cmd->resp_list[idx].status |= RECEIVED_ERR;	\
```
vishnuitta: this shouldn't cause any problem.. it may just delay the thread, until that gets timedout, which is waiting on this signal

vishnuitta: ok.. as I understood from you, setting 'status' can make rcomm_cmd get freed in other threads, and which is why this has been moved to the end.
Can you please add a comment for this, and, the precaution to be taken for further changes?

src/data_conn.c
```diff
@@ -210,22 +215,20 @@ handle_data_conn_error(replica_t *r)
 	spec = r->spec;
 
 	TAILQ_FOREACH(r1, &(spec->rq), r_next)
-		if (r1 == r)
+		if (r1 == r) {
+			update_spec_status= 1;
 			break;
-
-	if (r1 == NULL) {
```
vishnuitta: this condition is required here.
As comment from satbir, there are corner cases (update_replica_entry got timed out to check for some socket, but, it got created later) where 'r' may not be part of spec_rq list

src/data_conn.c
```diff
@@ -210,22 +215,20 @@ handle_data_conn_error(replica_t *r)
...
-		return -1;
-	}
-	TAILQ_REMOVE(&spec->rq, r, r_next);
+		}
 
 	if (r->state == ZVOL_STATUS_HEALTHY)
 		spec->healthy_rcount--;
 	else if (r->state == ZVOL_STATUS_DEGRADED)
 		spec->degraded_rcount--;
-	update_volstate(r->spec);
```
vishnuitta: why we need these changes here? there is lock on spec, so, removing from list should be fine.

src/data_conn.c
```diff
@@ -541,32 +545,33 @@ replica_thread(void *arg)
 	if (r_epollfd < 0) {
 		REPLICA_ERRLOG("epoll_create error for replica(%s:%d) "
 		    "errno(%d)\n", r->ip, r->port, errno);
-		return NULL;
+		goto initialize_error;
```
vishnuitta: this is good find..
let us do one more change:
- don't update 'replica' member variables till iofd is adding to epoll
- update 'replica' variables in the lock r_mtx when we are adding to epoll

src/replication.c
```diff
@@ -799,6 +816,16 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
 
 	spec->degraded_rcount++;
 	TAILQ_REMOVE(&spec->rwaitq, replica, r_waitnext);
+
+	rc = verify_replica_count(spec, replica);
```
vishnuitta: do the check before incrementing degraded_count, and check for '>=' condition in the verify_replica_count(). Also, change fn name like 'bool is_RF_replicas_connected()'.
Cleanup of created fds and replica thread need to be done. I think - that can be a tricky one.

vishnuitta: idea from you for checking before creating the thread is better.
Look at the diff like this:

[diff.log](https://github.com/openebs/istgt/files/2108033/diff.log)

vishnuitta (commented):
not related to this PR, 'shutdown' on an fd need to be called before calling 'close' to send the signal of closing to other side of fd.

src/data_conn.c
mynktl: ok

src/data_conn.c
mynktl: ok

src/data_conn.c
mynktl: ok

src/replication.c

src/replication.c

src/data_conn.c
mynktl: ok

src/data_conn.c
mynktl: ok

src/data_conn.c
mynktl: ok

src/replication.c

src/data_conn.c
mynktl: ok

src/data_conn.c
```diff
@@ -64,13 +64,27 @@ int replica_timeout = REPLICA_DEFAULT_TIMEOUT;
...
 			rcomm_cmd->resp_list[idx].status |= 		\
 			    RECEIVED_ERR;				\
 			pthread_cond_signal(_cond);			\
-		} else							\
+		} else {						\
 			rcomm_cmd->resp_list[idx].status |= 		\
 			    RECEIVED_ERR;				\
+			REPLICA_DEBUGLOG("error set for command(%lu)"	\
+			    " for replica(%lu)\n",			\
+			    rcomm_cmd->io_seq, r->zvol_guid);		\
```
vishnuitta: rcomm_cmd is again accessed after setting status, which can lead to crash

src/data_conn.c
```diff
@@ -626,6 +649,10 @@ replica_thread(void *arg)
 	ev.data.ptr = NULL;
 
 	MTX_LOCK(&r->r_mtx);
+	r->data_eventfd = r_data_eventfd;
```
vishnuitta: assign these vars after epoll_ctl of iofd is successful and within the lock

src/data_conn.c
```diff
@@ -635,19 +662,19 @@ replica_thread(void *arg)
 		close(r->mgmt_eventfd2);
 		r->mgmt_eventfd2 = -1;
 
-		if (r->epollfd > 0) {
+		if (r_epollfd > 0) {
 			/*
 			 * epoll_ctl may fail so we are ignoring return
 			 * value of epoll_ctl
 			 */
-			(void) epoll_ctl(r->epollfd, EPOLL_CTL_DEL,
```
vishnuitta: check r_data_eventfd > 0 also

mynktl: ok

src/data_conn.c
mynktl: ok

src/data_conn.c
mynktl: ok

src/data_conn.c
```diff
@@ -64,10 +64,24 @@ int replica_timeout = REPLICA_DEFAULT_TIMEOUT;
...
+			    RECEIVED_ERR;				\
+			pthread_cond_signal(_cond);			\
+		} else {						\
+			REPLICA_DEBUGLOG("error set for command(%lu)"	\
+			    " for replica(%lu)\n",			\
+			    rcomm_cmd->io_seq, r->zvol_guid);		\
+			rcomm_cmd->resp_list[idx].status |= 		\
+			    RECEIVED_ERR;				\
+		}							\
 		if (rcomm_cmd->state != CMD_EXECUTION_DONE)		\
```
vishnuitta: we shouldn't be using rcomm_cmd here, and, in next line also

mynktl: missed that one. i will remove that. thanks!

vishnuitta (approved):
changes are good

PR #41, opened 2018-06-15
[TA818] If replication factor is one, mark replica healthy during reb…

satbirchhikara:
If consistency factor is one, then we should be able to mark replica healthy as part of rebuild. This PR is to handle this case.

Unit testing is done for single replica case using mock replication code and setting replication factor as well as consistency factor to one in spec. Unit test case is automated.


vishnuitta (approved):
changes are good. Please resolve conflicts.

PR #43, opened 2018-06-17
STATS opcode implementation at target

vishnuitta:
This PR is to send STATS opcode to any healthy replica periodically, and handle its response.
Test cases are added for success/failure cases.


satbirchhikara (approved):
Looks good to me. Please go ahead.

PR #45, opened 2018-06-18
fixing error when istgt is started with no volume

mynktl:
Changes : 
- If `ISTGT` is started without any volume configuration then replication module will not accept connection from a replica.
- In `replication_test`, if it fails to make a connection with the target then it will exit with return value `1`.


src/istgt_lu.h
```diff
@@ -326,6 +328,7 @@ typedef struct istgt_lu_t {
 	ISTGT_LU_MAP map[MAX_LU_MAP];
 	int conns;
 #ifdef REPLICATION
+	uint64_t replica_id[MAXREPLICA];
```
vishnuitta: why is this required?

src/replication.c
```diff
@@ -1453,6 +1453,12 @@ accept_mgmt_conns(int epfd, int sfd)
                 }
                 MTX_UNLOCK(&specq_mtx);
 
+		if (!spec) {
+			REPLICA_ERRLOG("Spec is not configured\n");
+			close(mgmt_fd);
```
vishnuitta: you need to do 'shutdown' on fd also to inform other side of fd

src/replication_test.c
```diff
@@ -550,6 +550,11 @@ main(int argc, char **argv)
 					}
 					break;
 				}
+				if (count == 0) {
```
vishnuitta: are we hitting this when just 'close' is done on other side?

src/replication_test.c
```diff
@@ -614,6 +622,9 @@ main(int argc, char **argv)
 						} else if(((uint64_t)count < total_len - recv_len) && errno == EAGAIN) {
 							recv_len += count;
 							break;
+						} else if (count == 0) {
```
vishnuitta: this should be after initial 'if' check for count < 0..
same thing with next chunk of changes also

vishnuitta (commented):
testcase need to be added, like,
- bring up with no LU
- connect with replicas
- add LU section and do refresh
- connect should be done automatically

src/istgt_lu.h
mynktl: ok. My base branch was different. That's why this change is here. I will remove this change.

src/replication.c

src/replication_test.c
mynktl: No.  Actually, in `test_read_data`, we are not handling return value `0` from `read`. Due to this, have added this check in the caller of `test_read_data`.

src/replication_test.c
mynktl: ok

vishnuitta (approved):
changes are good

PR #46, opened 2018-06-18
Resolving istgtcontrol refresh issue

mynktl:
Changes : 
- If replication module is not initialized and the user executes CLI command
through `istgtcontrol` then `istgtcontrol` will return an error with a message
`replication module not initialized`.

- If `spec`'s status is not ready then `istgtcontrol status` will return volume `status` as `FAKE`.

Signed-off-by: mayank <mayank.patel@cloudbyte.com>-


src/istgt_lu_disk.c
```diff
@@ -6815,7 +6815,11 @@ istgt_lu_disk_status(ISTGT_LU_Ptr lu, int lun)
 		return -1;
 
 	MTX_LOCK(&spec->state_mutex);
-	status = spec->state;
+	MTX_LOCK(&spec->rq_mtx);
```
vishnuitta: why this change?

vishnuitta (commented):
testcase required

src/istgt_lu_disk.c
mynktl: With this change, In `istgtcontrol status`, if enough number of replicas are not connected to spec then we need to print volume status as `FAKE`.

vishnuitta (approved):
changes are good

PR #49, opened 2018-06-20
Removing file path from istgt.conf

mynktl:
Changes : 
Before this change, we were using dummy volume path in `istgt.conf`. With this change, we can remove volume path from `LUN Storage` option.
Now, In `istgt.conf` file, Storage option will have `size` and `block size` option only.
The format will be `LUN0 Storage size blocksize` _(`LUN0 Storage 1G 32k`)_


vishnuitta:
resolve the conflicts

mynktl:
@vishnuitta 
I have resolved confilct.

vishnuitta (approved):
changes are good

PR #50, opened 2018-06-29
[TA1304] unit tests for negative cases in target for network connections 

mynktl:
Changes : 
- Added error injection support in `istgt_integration` test utility. 
    As of now, error injection is added for below scenarios : 
       1. Closing management or data connection from replica
       2. Sending multiple responses for a single management command from replica
       3. Sending errored response for IOs from replica

- Code fix for the errors reported by error injection feature.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -415,16 +415,16 @@ trigger_rebuild(spec_t *spec)
 	replica_t *target_replica = NULL;
 	replica_t *healthy_replica = NULL;
 
+	if (spec->ready != true) {
+		REPLICA_NOTICELOG("Volume(%s) is not ready to accept IOs\n",
+		    spec->volname);
+		return;
+	}
+
 	if (spec->rebuild_in_progress == true) {
 		assert(spec->ready == true);
 		REPLICA_NOTICELOG("Rebuild is already in progress "
 		    "on volume(%s)\n", spec->volname);
-		return;
-	}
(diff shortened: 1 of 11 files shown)


src/replication.c
```diff
@@ -1479,6 +1478,10 @@ accept_mgmt_conns(int epfd, int sfd)
 	int mgmt_fd;
 	mgmt_event_t *mevent1, *mevent2;
 
+/*
```
vishnuitta: this is not required

src/replication.c
```diff
@@ -2253,6 +2259,8 @@ handle_mgmt_conn_error(replica_t *r, int sfd, struct epoll_event *events, int ev
 				if (r_ev == r) {
 					TAILQ_REMOVE(&r->spec->rwaitq, r, r_waitnext);
 					r->conn_closed++;
+					if (r->io_resp_hdr)
```
vishnuitta: set io_resp_hdr to NULL here, and also, in handle_data_conn_error() (with lock)

src/replication.c

src/replication.c

mynktl:
Hi @vishnuitta
I have addressed all review comments given by you. I have also enabled error injection for IOs and added comments for replica structure variables.

vishnuitta (approved):
changes are good.

PR #51, opened 2018-07-04
[TA482]  support for resizing volume in Target

mynktl:
Changes : 
- **Adding volume resize support in target.**
        Steps to resize volume:
            1. Update ISTGT config file (`istgt.conf`) with required volume size
            2. Execute `istgtcontrol refresh` command to update TARGET in-memory config. On error, its return status will be `1`.

- **Change in LUN config update path.**
     Previously LUN update has two different paths : 
          1.  By modifying LUN data structure if a client is logged in
          2.  By replacing old LUN data structure with a new one if no client is logged in.
  This LUN update path has a memory leak in case 2 (if no client is logged in). To make it simple and avoid memory leak, I have replaced addition/deletion path(case 2) with modification path.

- **Automated test to verify LUN resize functionality.**


src/replication.c
```diff
@@ -2764,6 +2713,215 @@ cleanup_deadlist(void *arg)
...
+	/* Wait for any ongoing mgmt commands which could block IOs. */
+	while (spec->quiesce == 1) {
+		MTX_UNLOCK(&spec->rq_mtx);
+		sleep(1);
+		MTX_LOCK(&spec->rq_mtx);
+	}
+
+	if (is_volume_healthy(spec) == false) {
+		REPLICA_ERRLOG("volume is not healthy..\n");
+		goto out;
```
satbirchhikara, line 2821: Memory leak of rcomm_mgmt_cmd if follow goto statement ?

mynktl: Yes. At this place, `rcomm_mgmt_cmd` hasn't been sent to replica so the caller should handle freeing rcomm_mgmt_cmd in this case.
 Thanks for catching this.

mynktl:
Thanks for reviewing changes @satbirchhikara  Sir.
I have updated this PR according to review comment, please have a look and let me know if any changes are required.

src/istgt_lu.c
```diff
@@ -1898,14 +1898,19 @@ istgt_lu_add_unit(ISTGT_Ptr istgt, CF_SECTION *sp)
...
+				int idx2 = 1;
 				if (lu->lun[i].type != ISTGT_LU_LUN_TYPE_NONE) {
 					ISTGT_ERRLOG("LU%d: duplicate LUN%d\n", lu->num, i);
 					goto error_return;
 				}
 				lu->lun[i].type = ISTGT_LU_LUN_TYPE_STORAGE;
 
-				size = istgt_get_nmval(sp, buf, j, 1);
-				rsz  = istgt_get_nmval(sp, buf, j, 2);
+#ifndef	REPLICATION
```
satbirchhikara, line 1908: I could not understand how it was working before replication. We were not using file in earlier code but now we are using it. Just want to know little more about "file" concept.

src/istgt_lu.c
```diff
@@ -2396,10 +2403,19 @@ istgt_lu_update_unit(ISTGT_LU_Ptr lu, CF_SECTION *sp)
 				ISTGT_ERRLOG("Unable to close Unauthorised connections for LU%d\n",lu->num);
 				goto error_out;
 			}
+
+			/*
+			 * If the client is logged in then we will skip
```
satbirchhikara: Do we know from maxtsih or igp_old that client is logged-in ? If "yes" then please amend comment so that future reader understand it clearly.

src/replication.c
```diff
@@ -2764,6 +2715,233 @@ cleanup_deadlist(void *arg)
...
+
+int
+istgt_lu_resize(spec_t *spec, uint64_t new_size, uint64_t old_size)
+{
+	rcommon_mgmt_cmd_t *rcomm_mgmt;
+	int rc = 0, rc1 = 0;
+
+	rcomm_mgmt = allocate_rcommon_mgmt_cmd(sizeof (new_size));
+	*(uint64_t *)rcomm_mgmt->buf = new_size;
+	rcomm_mgmt->buf_size = sizeof (new_size);
```
satbirchhikara, line 2915: We have already set rcomm_mgmt->buf_size inside allocate_rcommon_mgmt_cmd so we can remove this.

satbirchhikara (approved):
Apart from some minor comments, it looks good to me.

src/istgt_lu.c
mynktl: Earlier we were using dummy file for volume.

src/replication.c
mynktl: ok

src/istgt_lu.c
```diff
@@ -3236,24 +3281,31 @@ istgt_lu_reload_update(ISTGT_Ptr istgt)
...
 					MTX_LOCK(&lu->mutex);
 					if (lu->maxtsih > 1) {
-					/*	ISTGT_ERRLOG("update active LU%d: Name=%s, "
+						ISTGT_ERRLOG("update active LU%d: Name=%s, "
 						    "# of TSIH=%d\n",
-						    lu->num, lu->name, lu->maxtsih - 1);*/
+						    lu->num, lu->name, lu->maxtsih - 1);
+#ifdef	REPLICATION
+					} else {
+						ISTGT_ERRLOG("update LU%d: Name=%s\n", lu->num, lu->name);
```
vishnuitta, line 3289: why ERRLOG? above one also..

src/istgt_lu.c
```diff
@@ -3236,24 +3281,31 @@ istgt_lu_reload_update(ISTGT_Ptr istgt)
...
+							ISTGT_ERRLOG("Failed to update LUN\n");
+							return -1;
 						}
 						spec = (ISTGT_LU_DISK *)(lu->lun[0].spec);
 						MTX_LOCK(&spec->complete_queue_mutex);
 						rc = pthread_cond_signal(&spec->maint_cmd_queue_cond);
 						MTX_UNLOCK(&spec->complete_queue_mutex);
 						goto skip_lu;
+#ifndef	REPLICATION
 					} else {
```
vishnuitta, line 3309: this, probably, not required at all.. right?

PR #52, opened 2018-07-08
[TA1290]  Identifying the replicas that can connect to target

mynktl:
Changes : 
Added support for identifying replicas connecting to the target.
With this change,
1. The target will allow session from any replica until the number of connected(active or closed session) replicas reaches to replication factor limit. Once the replication factor limit reaches, Target will not allow any new replica to connect.
2. If the target has already an active session with a replica(`zvol_guid` is `12`) and same replica(with `zvol_guid` `12`) tries to make a new session with the target then the target will disconnect from the previous session and allow the new replica to connect.

Bug fix for replica cleanup path in data connection (in `handle_data_conn_error`).

### DIFF
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -293,6 +293,8 @@ handle_data_conn_error(replica_t *r)
 
 	MTX_LOCK(&r->r_mtx);
 	if (epoll_ctl(r->epollfd, EPOLL_CTL_DEL, r->iofd, NULL) == -1) {
+		shutdown(r->iofd, SHUT_RDWR);
+		r->iofd = -1;
 		MTX_UNLOCK(&r->r_mtx);
 		REPLICA_ERRLOG("epoll error for replica(%s:%d) iofd:%d "
 		    "err(%d)\n", r->ip, r->port, r->iofd, errno);
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -272,6 +272,15 @@ handle_data_conn_error(replica_t *r)
 	if (r1 == NULL) {
 		REPLICA_ERRLOG("replica %s %d not part of rqlist..\n",
 		    r->ip, r->port);
+		/*
+		 * mgmt thread will check mgmt_eventfd2 fd to see if
+		 * it needs to wait for replica thread or not.
+		 * At this stage, the replica hasn't been added to
+		 * spec's rqlist so we need to update mgmt_eventfd2 to -1
+		 * here So that mgmt_thread can skip replica thread check.
+		 */
+		if (r->mgmt_eventfd2 != -1)
+			r->mgmt_eventfd2 = -1;
 		MTX_UNLOCK(&spec->rq_mtx);
 		return -1;
 	}
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -292,14 +292,7 @@ handle_data_conn_error(replica_t *r)
 	MTX_UNLOCK(&spec->rq_mtx);
 
 	MTX_LOCK(&r->r_mtx);
-	if (epoll_ctl(r->epollfd, EPOLL_CTL_DEL, r->iofd, NULL) == -1) {
-		shutdown(r->iofd, SHUT_RDWR);
-		r->iofd = -1;
-		MTX_UNLOCK(&r->r_mtx);
-		REPLICA_ERRLOG("epoll error for replica(%s:%d) iofd:%d "
-		    "err(%d)\n", r->ip, r->port, r->iofd, errno);
-		return -1;
-	}
+	(void) epoll_ctl(r->epollfd, EPOLL_CTL_DEL, r->iofd, NULL);
 
 	fd = r->iofd;
 	r->iofd = -1;
--- a/src/replication.c
+++ b/src/replication.c
@@ -499,8 +499,20 @@ trigger_rebuild(spec_t *spec)
 	}
 }
 
+/*
+ * is_replica_newly_connected returns whether the connection is newly created
+ * from a familiar replica or newly connected from an unknown replica that
+ * target is waiting for.
+ * return value :
+ *	true  : if the connection is from a familiar replica (which does not
+ *		have an active session with the target) or from an unknown
+ *		replica for which target is waiting
+ *	false : if the connection is from a familiar replica which is already
+ *		connected to target or target is not waiting for any new
+ *		replica.
+ */
 static bool
-is_replica_new_or_familar(spec_t *spec, replica_t *new_replica)
+is_replica_newly_connected(spec_t *spec, replica_t *new_replica)
 {
 	known_replica_t *kr = NULL;
 	int familiar_replicas = 0;
@@ -827,11 +839,11 @@ int
 update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
 {
 	int rc;
-	zvol_io_hdr_t *rio_hdr;
+	zvol_io_hdr_t *rio_hdr = NULL;
 	pthread_t r_thread;
 	zvol_io_hdr_t *ack_hdr;
 	mgmt_ack_t *ack_data;
-	zvol_op_open_data_t *rio_payload;
+	zvol_op_open_data_t *rio_payload = NULL;
 	int i;
 
 	ack_hdr = replica->mgmt_io_resp_hdr;	
@@ -853,6 +865,16 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
 
 	replica->pool_guid = ack_data->pool_guid;
 	replica->zvol_guid = ack_data->zvol_guid;
+
+	MTX_LOCK(&spec->rq_mtx);
+	if (!is_replica_newly_connected(spec, replica)) {
+		MTX_UNLOCK(&spec->rq_mtx);
+		REPLICA_ERRLOG("replica(ip:%s port:%d "
+		    "guid:%lu) is not permitted to connect\n", replica->ip, replica->port,
+		    replica->zvol_guid);
+		goto replica_error;
+	}
+	MTX_UNLOCK(&spec->rq_mtx);
 
 	replica->spec = spec;
 	replica->io_resp_hdr = (zvol_io_hdr_t *) malloc(sizeof (zvol_io_hdr_t));
@@ -929,13 +951,6 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
 		goto replica_error;
 	}
 
-	if (!is_replica_new_or_familar(spec, replica)) {
-		MTX_UNLOCK(&spec->rq_mtx);
-		REPLICA_ERRLOG("replica(ip:%s port:%d "
-		    "guid:%lu) is not permitted to connect\n", replica->ip, replica->port,
-		    replica->zvol_guid);
-		goto replica_error;
-	}
 	MTX_UNLOCK(&spec->rq_mtx);
 
 	rc = pthread_create(&r_thread, NULL, &replica_thread,
@@ -946,8 +961,10 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
 replica_error:
 		replica->iofd = -1;
 		close(iofd);
-		free(rio_hdr);
-		free(rio_payload);
+		if (rio_hdr)
+			free(rio_hdr);
+		if (rio_payload)
+			free(rio_payload);
 		return -1;
 	}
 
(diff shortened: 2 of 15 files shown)


src/data_conn.c
```diff
@@ -293,6 +293,8 @@ handle_data_conn_error(replica_t *r)
 
 	MTX_LOCK(&r->r_mtx);
 	if (epoll_ctl(r->epollfd, EPOLL_CTL_DEL, r->iofd, NULL) == -1) {
+		shutdown(r->iofd, SHUT_RDWR);
```
satbirchhikara: If epoll_ctl system call got an error, it implies that something weird happened in the system. In such case better to get rid-off this replica. Why we do by returning from here ?

src/replication.c
```diff
@@ -488,6 +497,57 @@ trigger_rebuild(spec_t *spec)
...
+			break;
+		}
+	}
+
+	if (!found && connected_replica < spec->replication_factor) {
+		kr = malloc(sizeof (known_replica_t));
+		kr->zvol_guid = new_replica->zvol_guid;
+		kr->is_connected = true;
+		ret = true;
+		TAILQ_INSERT_TAIL(&spec->identified_replica, kr, next);
```
satbirchhikara: I can see that we are inserting replica in to "identified_replica" list but we are not removing anything from this list anywhere in code. Is it by design or we missed freeing it ?

src/replication.c
```diff
@@ -1893,23 +1978,25 @@ empty_mgmt_q_of_replica(replica_t *r)
...
+			switch (mgmt_cmd->io_hdr->opcode) {
+				case ZVOL_OPCODE_SNAP_CREATE:
+					handle_snap_create_resp(r, mgmt_cmd);
+					break;
+				case ZVOL_OPCODE_PREPARE_FOR_REBUILD:
+					handle_prepare_for_rebuild_resp(r->spec,
+					    mgmt_cmd->io_hdr, NULL, mgmt_cmd);
+					break;
+				case ZVOL_OPCODE_STATS:
+					update_spec_stats(r->spec, mgmt_cmd->io_hdr,
```
satbirchhikara: How about rename to "handle_update_spec_stats" just for symmetry ??

src/replication.c
```diff
@@ -488,6 +497,57 @@ trigger_rebuild(spec_t *spec)
...
+				ret = true;
+			}
+			break;
+		}
+	}
+
+	if (!found && connected_replica < spec->replication_factor) {
+		kr = malloc(sizeof (known_replica_t));
+		kr->zvol_guid = new_replica->zvol_guid;
+		kr->is_connected = true;
```
satbirchhikara: What if replica is already there in identified_replica list but not connected(is_connected == false) where would we set it to true when we move this replica to ready_queue (spec->rq) ? Connected_replica should be count of replicas which are actually connected i.e. kr->is_connected should be true for those replicas otherwise we are just counting stale replica.

src/data_conn.c
mynktl: yes. This is replica removal path from the target, So we can ignore error here.

src/replication.c
mynktl: Yes.`is_connected` should be set to `true` if replica re-connects.

src/replication.c
mynktl: yes. It is by design. With this, we are allowing first `replication_factor` number of replicas to connect.

src/replication.c

mynktl (commented):
Thanks for reviewing changes @satbirchhikara Sir.
I have tried to address all review comments, please have a look and let me know if any changes are required.

src/replication.c
```diff
@@ -381,13 +384,17 @@ start_rebuild(void *buf, replica_t *replica, uint64_t data_len)
...
+
 	mgmt_cmd = malloc(sizeof (mgmt_cmd_t));
+	memset(mgmt_cmd, 0, sizeof (mgmt_cmd_t));
 	BUILD_REPLICA_MGMT_HDR(rmgmtio, mgmt_opcode, data_len);
 
 	mgmt_cmd->io_hdr = rmgmtio;
 	mgmt_cmd->io_bytes = 0;
 	mgmt_cmd->data = buf;
 	mgmt_cmd->mgmt_cmd_state = WRITE_IO_SEND_HDR;
+	mgmt_cmd->rcomm_mgmt = NULL;
```
satbirchhikara: Since we are zeroing this memory, we do not need code at line 394 and at 397.

src/replication.c
```diff
@@ -933,6 +1003,7 @@ send_replica_handshake_query(replica_t *replica, spec_t *spec)
 	int ret = 0;
 
 	mgmt_cmd = malloc(sizeof(mgmt_cmd_t));
+	memset(mgmt_cmd, 0, sizeof (mgmt_cmd_t));
```
satbirchhikara: Once we zero out this memory, we do not need code at line 1016 and 1019. We can remove those lines.

src/replication.c
```diff
@@ -998,6 +1069,7 @@ send_replica_snapshot(spec_t *spec, replica_t *replica, uint64_t io_seq,
...
```
satbirchhikara: Same comment goes here too.

src/replication.c
```diff
@@ -1230,6 +1302,7 @@ send_replica_query(replica_t *replica, spec_t *spec, zvol_op_code_t opcode)
 	mgmt_cmd_t *mgmt_cmd;
 
 	mgmt_cmd = malloc(sizeof(mgmt_cmd_t));
+	memset(mgmt_cmd, 0, sizeof (mgmt_cmd_t));
```
satbirchhikara: Same goes here.

satbirchhikara (approved):
Some minor comments given. Please have a look and let me know if you need more clarification.

src/replication.c

src/replication.c

src/replication.c

src/replication.c

src/data_conn.c
```diff
@@ -292,12 +301,7 @@ handle_data_conn_error(replica_t *r)
 	MTX_UNLOCK(&spec->rq_mtx);
 
 	MTX_LOCK(&r->r_mtx);
-	if (epoll_ctl(r->epollfd, EPOLL_CTL_DEL, r->iofd, NULL) == -1) {
```
vishnuitta, line 295: why this change?

src/replication.c
```diff
@@ -500,6 +499,59 @@ trigger_rebuild(spec_t *spec)
 	}
 }
 
+static bool
+is_replica_familiar(spec_t *spec, replica_t *new_replica)
+{
+	known_replica_t *kr = NULL;
+	int familiar_replicas = 0;
+	bool ret = false, found = false;
```
vishnuitta: its better to change the name of 'ret' to 'newly_connected'

src/replication.c
```diff
@@ -500,6 +499,59 @@ trigger_rebuild(spec_t *spec)
...
+	 * then the target will disconnect from the previous session and
+	 * allow a new session
+	 */
+	if (found && !ret) {
+		TAILQ_FOREACH(old_replica, &spec->rq, r_next) {
+			if (old_replica->zvol_guid == new_replica->zvol_guid) {
+				MTX_UNLOCK(&spec->rq_mtx);
+				REPLICA_LOG("Disconnecting replica(%lu)'s old "
+				    "session\n", old_replica->zvol_guid);
+				handle_mgmt_conn_error(old_replica, 0, NULL, 0);
```
vishnuitta: during this, what happens for any pending events received in 'init_replication' thread?

src/replication.c
```diff
@@ -500,6 +499,59 @@ trigger_rebuild(spec_t *spec)
...
+		ret = true;
+		TAILQ_INSERT_TAIL(&spec->identified_replica, kr, next);
+	}
+
+	/*
+	 * If the target has already an active session with this replica
+	 * then the target will disconnect from the previous session and
+	 * allow a new session
+	 */
+	if (found && !ret) {
```
vishnuitta: let us continue the old one, and error out the new connection due to complexities in looking for any pending IOs / pending epoll signals.
Is there any reason why we should consider new connection?

src/replication.c
```diff
@@ -873,13 +925,21 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
 	}
 
 	MTX_LOCK(&spec->rq_mtx);
-	if(is_rf_replicas_connected(spec, replica)) {
+	if (is_rf_replicas_connected(spec, replica)) {
 		MTX_UNLOCK(&spec->rq_mtx);
 		ISTGT_ERRLOG("failed to verify replica(ip:%s port:%d "
```
vishnuitta: seems like, this log need to be fixed (not related to this PR)

src/replication.c
```diff
@@ -873,13 +925,21 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
...
 		    "guid:%lu)\n", replica->ip, replica->port,
 		    replica->zvol_guid);
 		goto replica_error;
 	}
+
+	if (!is_replica_familiar(spec, replica)) {
```
vishnuitta: 'is_replica_new_or_familar' may be good?

src/data_conn.c
mynktl: Here, `handle_data_conn_error` was returning an error if `epoll_ctl` fails which may block management thread which will be waiting for `replica_thread`.
Since it is a cleanup path, we can ignore `epoll_ctl` return value and continue further cleanup.

src/replication.c

src/replication.c

src/replication.c

src/replication.c
mynktl: Good find!
We will not allow a new connection due to complexity. At this point, we won't be able to know any pending events on old replica session.

src/replication.c
mynktl: let us continue the old one, and error out the new connection due to complexities in looking for any pending IOs / pending EPOLL signals.
>Good find! We can avoid new connection if we set keepalive for TCP socket. I have created an issue(**US2827**) for the same.

Is there any reason why we should consider a new connection?
>If the system(replica node) crash happens then target won't get any event regarding replica disconnect. Once system/replica comes up then it will try to make a new connection. Now, the target will have two sessions from the same replica.
As of now, the old session will take some time for disconnect event since we are not setting keepalive for a socket. Due to this, we are considering the new connection.

src/replication.c
```diff
@@ -500,6 +499,54 @@ trigger_rebuild(spec_t *spec)
...
+
+	/*
+	 * If the target has already an active session with this replica
+	 * then the target will not allow new session
+	 */
+	if (found && !newly_connected) {
+		REPLICA_ERRLOG("replica(%lu) has already an active session\n",
+		    kr->zvol_guid);
+		/*
+		 * handle_mgmt_conn_error will update identified_replica status
```
vishnuitta, line 552: thats a very good explanation @mynktl ..
I'm thinking whether same scenario (i.e., another connection from same replica) can apply to other error scenarios in update_replica_entry after setting zvol_guid.
We need to move this check for is_replica_new_or_familiar right after setting zvol_guid in update_replica_entry.
Sorry to ask one more change along with above thing: can you change is_replica_new_or_familiar to is_replica_newly_connected and with some comments saying 'This function returns whether connection is newly created from a familiar replica or newly connected from an unknown replica that target is waiting for'?

src/replication.c
```diff
@@ -2617,6 +2676,7 @@ initialize_volume(spec_t *spec, int replication_factor, int consistency_factor)
 	TAILQ_INIT(&spec->rcommon_waitq);
 	TAILQ_INIT(&spec->rq);
 	TAILQ_INIT(&spec->rwaitq);
+	TAILQ_INIT(&spec->identified_replica);
```
vishnuitta, line 2703: should we destroy the list during destroy_volume?

mynktl: PR which contains destroy_volume function is not yet merged. I thought of handling destroying identified_replica in that PR #54 once this PR gets merged.

src/replication.c
mynktl: ok

src/replication.c
```diff
@@ -2365,6 +2434,13 @@ handle_mgmt_conn_error(replica_t *r, int sfd, struct epoll_event *events, int ev
...
 		r->spec->rebuild_in_progress = false;
 	}
+
+	TAILQ_FOREACH(kr, &r->spec->identified_replica, next) {
+		if (kr->zvol_guid == r->zvol_guid) {
+			kr->is_connected = false;
+			break;
+		}
+	}
 	MTX_UNLOCK(&r->spec->rq_mtx);
```
vishnuitta, line 2452: we can add an ASSERT for debug build if kr is NULL

vishnuitta: ok.. may be, not required..
if too many replicas gets connected, there are chances that kr may be NULL

vishnuitta (approved):
changes are good

PR #53, opened 2018-07-08
[TA1699] Error handling for mutiple responses for a mgmt command from a replica

mynktl:
Change : 
1. Since target doesn't allow parallel execution of mgmt command, If the target receives multiple responses for a mgmt command from the same replica then the target will disconnect that replica.

2. Fixes regarding memory leak in replica interface and integration test (only a few leaks are taken care, rest will be resolved in separate user story)

3. Few corner fixes in rebuild interface.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -391,10 +391,8 @@ start_rebuild(void *buf, replica_t *replica, uint64_t data_len)
 	BUILD_REPLICA_MGMT_HDR(rmgmtio, mgmt_opcode, data_len);
 
 	mgmt_cmd->io_hdr = rmgmtio;
-	mgmt_cmd->io_bytes = 0;
 	mgmt_cmd->data = buf;
 	mgmt_cmd->mgmt_cmd_state = WRITE_IO_SEND_HDR;
-	mgmt_cmd->rcomm_mgmt = NULL;
 
 	MTX_LOCK(&replica->r_mtx);
 	TAILQ_INSERT_TAIL(&replica->mgmt_cmd_queue, mgmt_cmd, mgmt_cmd_next);
--- a/src/replication.c
+++ b/src/replication.c
@@ -250,6 +250,7 @@ enqueue_prepare_for_rebuild(spec_t *spec, replica_t *replica,
 	uint64_t data_len = strlen(spec->volname) + 1;
 
 	mgmt_cmd = malloc(sizeof (mgmt_cmd_t));
+	memset(mgmt_cmd, 0, sizeof (mgmt_cmd_t));
 	BUILD_REPLICA_MGMT_HDR(rmgmtio, opcode, data_len);
 
 	mgmt_cmd->io_hdr = rmgmtio;
@@ -298,6 +299,8 @@ send_prepare_for_rebuild_or_trigger_rebuild(spec_t *spec,
 	mgmt_ack_t *mgmt_data;
 	replica_t *replica;
 	struct rcommon_mgmt_cmd *rcomm_mgmt;
+
+	ASSERT(MTX_LOCKED(&spec->rq_mtx));
 
 	spec->target_replica = target_replica;
 	spec->rebuild_in_progress = true;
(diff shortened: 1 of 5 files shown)


src/replication.c
```diff
@@ -1893,23 +1919,25 @@ empty_mgmt_q_of_replica(replica_t *r)
...
+					    mgmt_cmd->io_hdr, NULL, mgmt_cmd);
+					break;
+				case ZVOL_OPCODE_STATS:
+					update_spec_stats(r->spec, mgmt_cmd->io_hdr,
+					    NULL);
+					break;
+				default:
+					break;
+			}
+			REPLICA_NOTICELOG("mgmt command(%d) failed for replica(%lu)\n",
```
vishnuitta, line 1940: there seems to be some code merge issue here

mynktl: No. It is not a merge issue. Here, we are skipping mgmt command handling if `cmd_completed` is set to `true`.

satbirchhikara: I did not see change w.r.t "Disconnect if multiple mgmt commands from same replica" fix.

mynktl: in `read_io_resp`, we are trying to read for a response header once we get data-response or complete response from a replica. If read for a response header returns success then we set `donecount` to `-1` in `CHECK_AND_ADD_BREAK_IF_PARTIAL` and return it to caller.

src/replication.c
```diff
@@ -381,13 +384,17 @@ start_rebuild(void *buf, replica_t *replica, uint64_t data_len)
...
+
 	mgmt_cmd = malloc(sizeof (mgmt_cmd_t));
+	memset(mgmt_cmd, 0, sizeof (mgmt_cmd_t));
 	BUILD_REPLICA_MGMT_HDR(rmgmtio, mgmt_opcode, data_len);
 
 	mgmt_cmd->io_hdr = rmgmtio;
 	mgmt_cmd->io_bytes = 0;
 	mgmt_cmd->data = buf;
 	mgmt_cmd->mgmt_cmd_state = WRITE_IO_SEND_HDR;
+	mgmt_cmd->rcomm_mgmt = NULL;
```
vishnuitta: not required.. already memset to 0

src/replication.c
```diff
@@ -250,6 +250,7 @@ enqueue_prepare_for_rebuild(spec_t *spec, replica_t *replica,
 	uint64_t data_len = strlen(spec->volname) + 1;
 
 	mgmt_cmd = malloc(sizeof (mgmt_cmd_t));
+	memset(mgmt_cmd, 0, sizeof (mgmt_cmd_t));
```
vishnuitta, line 257: this is a good find. List inside mgmt_cmd is not initialized at many places.

src/replication.c

vishnuitta (approved):
changes are good. Will do review once again.

src/replication.c
```diff
@@ -1897,23 +1921,25 @@ empty_mgmt_q_of_replica(replica_t *r)
...
+					update_spec_stats(r->spec, mgmt_cmd->io_hdr,
+					    NULL);
+					break;
+				default:
+					break;
+			}
+			REPLICA_NOTICELOG("mgmt command(%d) failed for replica(%lu)\n",
+			    mgmt_cmd->io_hdr->opcode, r->zvol_guid);
 		}
-		REPLICA_NOTICELOG("mgmt command(%d) failed for replica(%lu)\n",
```
vishnuitta, line 1915: lets add a LOG here

mynktl: Here we have two types of management command...
1. Commands which were queued in mgmt queue and haven't received any response.
2. A command for which we have already received a response but due to multiple responses `read_io_resp` has reported an error and didn't dequeue it from mgmt command queue.

For 1st type, we are logging error in `empty_mgmt_q_of_replica` and for 2nd type, we are logging it as `protocol error` in `read_io_resp`.

PR #58, opened 2018-07-24
[TA2090] Sync command from iSCSI controller should not pass length.

satbirchhikara:
uZFS not expecting any length with SYNC command. Fixing same. Found as part of GKE integration testing.
Also, modified CI to use ext4 instead of ext2 during testing.


vishnuitta (approved):
changes are good.

PR #60, opened 2018-08-02
[US2036] selecting healthy replica with minimal in-flight read IOs to…

mynktl:
Changes : 

- selecting healthy replica with minimal in-flight read IOs to serve read IOs. With this approach, Workload for the read IOs will be shared across all healthy replicas, which will reduce latency for the read IOs.

FIO latency-throughput results for the read IOS with/without these changes.
**Without these changes:** 
Latency test (Latency: 39.34 ms, Iops: 813)
Throughput test (Latency: 41.37 ms, Iops: 3092)

**With these changes:** 
Latency test (Latency: 31.46 ms, Iops: 1016)
Throughput test (Latency: 33.38 ms, Iops: 3831)

_ConsistencyFactor : 2_
_Test run time: 30 second_
_FIO Block size: 4k_
_IO-depth: 32_
_Thread: 1 (for latency test) 4(for throughput test)_


src/replication.c
```diff
@@ -2206,6 +2244,9 @@ replicate(ISTGT_LU_DISK *spec, ISTGT_LU_CMD_Ptr cmd, uint64_t offset, uint64_t n
 		if (cmd_write)
 			__sync_fetch_and_add(&replica->replica_inflight_write_io_cnt, 1);
 
+		if (cmd_read)
```
satbirchhikara: Would be great if we rewrite it like.....
if ()
else if()

Also please put read check first since reads are 60%-70% of total IO

src/replication.c
```diff
@@ -2141,14 +2141,17 @@ int64_t
...
 {
 	int rc = -1, i;
-	bool cmd_write = false;
-	replica_t *replica;
+	bool cmd_write = false, cmd_read = false;
+	replica_t *replica, *last_replica = NULL;
 	rcommon_cmd_t *rcomm_cmd;
 	rcmd_t *rcmd = NULL;
 	int iovcnt = cmd->iobufindx + 1;
 	bool cmd_sent = false;
```
satbirchhikara: Can we rename cmd_sent to replica_choosen ?

satbirchhikara (approved):
Looks good to me.

src/replication.c
```diff
@@ -2225,6 +2266,12 @@ replicate(ISTGT_LU_DISK *spec, ISTGT_LU_CMD_Ptr cmd, uint64_t offset, uint64_t n
 		if (rc) {
 			if (rc == 1) {
 				rc = cmd->data_len = rcomm_cmd->data_len;
+			} else if (rcomm_cmd->opcode == ZVOL_OPCODE_READ &&
+			    rcomm_cmd->copies_sent == 1) {
+				rcomm_cmd->copies_sent = 0;
+				memset(rcomm_cmd->resp_list, 0, sizeof (rcomm_cmd->resp_list));
+				MTX_LOCK(&spec->rq_mtx);
```
vishnuitta, line 2350: should we reset cmd_sent local variable also?

vishnuitta: IO may be going on with few other replicas which are slow, and, may overwrite rcomm_cmd->resp_list with the index given in its rcmd_t (replica_cmd). This can lead to issues.

src/replication.c
```diff
@@ -2141,14 +2141,17 @@ int64_t
...
+	replica_t *replica, *last_replica = NULL;
 	rcommon_cmd_t *rcomm_cmd;
 	rcmd_t *rcmd = NULL;
 	int iovcnt = cmd->iobufindx + 1;
 	bool cmd_sent = false;
 	struct timespec abstime, now;
 	int nsec, err_num = 0;
+	int skip_count = 0;
+	uint64_t num_read_ios = 0;
+	int64_t read_io_diff = 0;
```
vishnuitta: this is not in use?

src/replication.c
vishnuitta: num_read_ios local variable also need to reset

src/replication.c

src/replication.c
mynktl: yes. I will remove this

src/replication.c

src/replication.c
mynktl: I will reset `cmd_sent` and `num_read_ios` variable.
Regarding index in `rcomm_cmd->resp_list`, we are doing retry only if we are sending read command to a single replica.

src/replication.c
```diff
@@ -2271,21 +2288,47 @@ replicate(ISTGT_LU_DISK *spec, ISTGT_LU_CMD_Ptr cmd, uint64_t offset, uint64_t n
 			 */
 			if (replica->state == ZVOL_STATUS_DEGRADED)
 				continue;
-			else
-				cmd_sent = true;
+			else {
+				inflight_read_ios = replica->replica_inflight_read_io_cnt;
```
vishnuitta, line 2292: this function is becoming big.. its better to modularize it.

vishnuitta, line 2292: use sync_fetch_and_add rather than directly using it, to avoid any data races

vishnuitta: this being 64-bit aligned, we are fine with 64 bit machines.

vishnuitta (approved):
changes are good. Automated test cases can be added in next PR.

PR #61, opened 2018-08-02
[DE80] Fixing read consistency test in test_istgt script

mynktl:
Changes: 
Due to invalid length in `write_data` API call, read consistency was failing. This change includes the fix for the same.


vishnuitta (approved):
changes are good

PR #63, opened 2018-08-09
[US2082] Enabling SPC2 and setting default IO timeout in handshake

vishnuitta:
Currently, istgt code advertises that it supports SPC3 reservations. But, in cStor volumes, replicas doesn't support SPC3.

This PR is to disable SPC3 and enable SPC2 reservations while advertising to client.
Update:
Setting Standard Inquiry parameters to SPC2 is limiting initiator not to read BLOCK_LIMITS vpd page.
This is making initiator to send write IOs that are more than 1MB, where as, istgt is configured to throw error if write IO size is more than 1MB.
Had set the standard inquiry parameters to SPC3 (as usual), and started returning error for RESERVE_IN and RESERVE_OUT CDBs.

Also, currently, istgt sends 10*60 seconds as IO time out in its handshake message. This PR changes hardcoded value to 3 times of configured replica_timeout value.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -902,7 +902,7 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
 
 	rio_payload = (zvol_op_open_data_t *) malloc(
 	    sizeof (zvol_op_open_data_t));
-	rio_payload->timeout = (10 * 60);
+	rio_payload->timeout = (3 * replica_timeout);
 	rio_payload->tgt_block_size = spec->blocklen;
 	strncpy(rio_payload->volname, spec->volname,
 	    sizeof (rio_payload->volname));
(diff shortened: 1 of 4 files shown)


src/replication.c
```diff
@@ -902,7 +902,7 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
 
 	rio_payload = (zvol_op_open_data_t *) malloc(
 	    sizeof (zvol_op_open_data_t));
-	rio_payload->timeout = (10 * 60);
+	rio_payload->timeout = (3 * replica_timeout);
```
satbirchhikara, line 905: Where is this replica_timeout defined ?

satbirchhikara (approved):
LGTM

mynktl (approved):

mynktl (approved):

src/replication.c
vishnuitta: this comes either through istgt cmd line options or by default initialized to 200 seconds in the code or set to 10 seconds in istgt_integration code during run time.

payes (approved):

PR #66, opened 2018-08-15
[TA2388] quiescing sync IO during snapshot command

mynktl:
Changes : 
- quiescing sync IO during snapshot command
- Changing ISTGT logging from syslog to fprintf (stderr)
- Changing rebuild wait time for the newly added replica to 2 times of replica_timeout
- Fixing deadlock in update_replica_entry


vishnuitta (commented):
Changes are good.
- Instead of using INCREASE../DECREASE.. kind of macros, can we change them to INCREMENT../DECREMENT..?
- Lets leave build_rcmd as it is, and leave to its callers to update the inflight count
- automated test cases are required

src/replication_log.h
```diff
@@ -0,0 +1,51 @@
...
+	LOG_LEVEL_INFO,
+	LOG_LEVEL_DEBUG,
+	LOG_LEVEL_MAX,
+};
+
+int replication_log_level;
+
+#define REPLICA_LOG(fmt, ...)						\
+{									\
+	if (replication_log_level <= LOG_LEVEL_INFO)			\
```
vishnuitta: these conditions should be reverse if I'm not wrong

mynktl: yes

vishnuitta (approved):
changes are good

PR #67, opened 2018-08-16
[US2543] Removing debug-symbols files for cStor target

mynktl:
Changes : 
- Removing `*.full` binary in build since it is not needed.
- Removing debug-symbol files from the build
- Making final executables(`istgt`, `istgtcontrol`) unstripped.


vishnuitta (approved):
changes are good

PR #68, opened 2018-08-17
Added timestamp for logs

vishnuitta:
Currently, istgt doesn't have timestamp in its logs. This PR is to add timestamp in the istgt logs.


mynktl (approved):
changes are good.

PR #71, opened 2018-08-22
[TA2578]  command for ISTGTCONTROL to list replica and its stats

mynktl:
Changes : 
- Added `REPLICA` command in `ISTGTCONTROL` to list replica and its stats.
Command : `istgtcontrol REPLICA <volname>`
Note: `volname` is an optional argument. If specified then it will list replica for that volume only else it will list all volume and its replica.
Sample output : 
```jsonc
{
  "Replica status": [
    {
      "vol1": [
        {
          "replica": 6063,
          "status": "DEGRADED",
          "checkpointed_io_seq": 1000,
          "inflight_read": 0,
          "inflight_write": 0,
          "inflight_sync": 0,
          "connected since(in seconds)": 64
        },
        {
          "replica": 6061,
          "status": "DEGRADED",
          "checkpointed_io_seq": 1000,
          "inflight_read": 0,
          "inflight_write": 0,
          "inflight_sync": 0,
          "connected since(in seconds)": 63
        },
        {
          "replica": 6062,
          "status": "DEGRADED",
          "checkpointed_io_seq": 1000,
          "inflight_read": 0,
          "inflight_write": 0,
          "inflight_sync": 0,
          "connected since(in seconds)": 62
        }
      ]
    }
  ]
}
```


src/replication.c
```diff
@@ -1370,6 +1371,101 @@ int istgt_lu_create_snapshot(spec_t *spec, char *snapname, int io_wait_time, int
...
+			json_object_object_add(j_spec, spec->volname, j_replica);
+			json_object_array_add(j_stats, j_spec);
+		}
+	}
+
+	MTX_UNLOCK(&specq_mtx);
+
+	j_obj = json_object_new_object();
+	json_object_object_add(j_obj, "Replica status", j_stats);
+	json_string = json_object_to_json_string_ext(j_obj,
```
vishnuitta, line 1462: how about freeing this string?

vishnuitta (approved):
changes are good

PR #78, opened 2018-09-11
Avoid iterating through a replica's blocked_queue if the next command in the blocked_queue is in a blocked state.

mynktl:
Changes : 
- Avoid iterating through a replica's blocked_queue if the next command in the blocked_queue is in a blocked state.
   As of now, we are iterating through a replica's blocked_cmd_queue and if a command is not blocked then we are enqueuing it to the ready_queue. With this approach, There are chances that the command sequence will be different in the replica's ready_queue compare to the replica's cmd_queue. To avoid this, iterating through a blocked_queue should be stopped if a command from blocked_queue is in the blocked state.


vishnuitta (approved):
changes are good

PR #80, opened 2018-09-18
[TA2332] feat(snap_rebuild) : allow snapshot to be taken

pawanpraka1:
if number of healthy replicas greater or equal to consistency factor

Signed-off-by: Pawan <pawanprakash101@gmail.com>

### DIFF
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -732,12 +732,8 @@ initialize_error:
 			if (ptr != NULL)
 				goto exit;
 
-			MTX_LOCK(&r->r_mtx);
-			if (r->disconnect_conn) {
-				MTX_UNLOCK(&r->r_mtx);
+			if (r->disconnect_conn)
 				goto exit;
-			}
-			MTX_UNLOCK(&r->r_mtx);
 
 			if (events[i].events & (EPOLLERR | EPOLLRDHUP | EPOLLHUP))
 				goto exit;
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -712,6 +712,12 @@ initialize_error:
 			goto exit;
 		}
 
+		if (r->disconnect_conn) {
+			ret = -1;
+			REPLICA_ERRLOG("replica disconnected(%s:%d)\n", r->ip, r->port);
+			goto exit;
+		}
+
 		for (i = 0; i < nfds; i++) {
 			fd = events[i].data.fd;
 			if ((fd == r_data_eventfd) || (fd == r_mgmt_eventfd)) {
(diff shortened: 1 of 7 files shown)


satbirchhikara (commented):
Looks good to me still want Vitta sir to look in to it.
-Satbir

src/data_conn.c
```diff
@@ -731,6 +732,13 @@ replica_thread(void *arg)
 			if (ptr != NULL)
 				goto exit;
 
+			MTX_LOCK(&r->r_mtx);
+			if (r->disconnect_conn) {
```
vishnuitta: currently, normal IO code path of replica doesn't have locks.. its better to avoid this lock..
what if disconnect_conn is set to 1 after checking here?

src/replication.c
```diff
@@ -1363,12 +1375,21 @@ int istgt_lu_create_snapshot(spec_t *spec, char *snapname, int io_wait_time, int
 	rcomm_mgmt->caller_gone = 1;
 	if (rcomm_mgmt->cmds_sent == (rcomm_mgmt->cmds_succeeded + rcomm_mgmt->cmds_failed)) {
 		free_rcomm_mgmt = 1;
-		if ((rcomm_mgmt->cmds_succeeded == spec->replication_factor) && (rcomm_mgmt->cmds_failed == 0)) {
+		if (rcomm_mgmt->cmds_succeeded >= cf) {
```
vishnuitta: this can be moved out of the 'if' condition (cmds_sent == cmds_succeeded + cmds_failed)

src/replication.c
```diff
@@ -1189,6 +1189,24 @@ send_replica_snapshot(spec_t *spec, replica_t *replica, uint64_t io_seq,
...
+    zvol_op_code_t opcode)
+{
+	mgmt_cmd_t *mgmt_cmd;
+
+	MTX_LOCK(&replica->r_mtx);
+	TAILQ_FOREACH(mgmt_cmd, &replica->mgmt_cmd_queue, mgmt_cmd_next) {
+		if (mgmt_cmd->io_hdr->io_seq == io_seq &&
+		    mgmt_cmd->io_hdr->opcode == opcode) {
+			inform_mgmt_conn(replica);
+			break;
```
vishnuitta, line 1203: we may need to continue the loop

src/data_conn.c
pawanpraka1: hmmm, will remove the lock. if disconnect_conn is set to 1 after the check we will process one event, it's the responsibility of the the person who is setting this to one, to make sure setting this before getting any event.

src/replication.c
pawanpraka1: done.

src/replication.c
pawanpraka1: not needed, as disconnect_conn is replica level flag. So we can break from here.

vishnuitta (approved):
changes are good. Test automation can be taken with different task.

PR #81, opened 2018-09-27
fix(build): fetch the zfs header from the correct branch

kmova:
autoreconf is always fetching the zfs header file from zfs repo
master branch. This make it incompatible to build on released
branches if there has been a change to the header file on the master.

Added script to determine the zfs branch from which the header
file should be fetched.
- If tagged build, fetch from the corresponding tag
- If master (replication), fetch from the zfs master
- If non-master, fetch from corresponding zfs branch if it exists
  or use the zfs master

### DIFF
--- a/fetch-zfs-branch.sh
+++ b/fetch-zfs-branch.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+set -e
+
+if [ ! -z ${TRAVIS_TAG} ];
+then
+  ZFS_BUILD_BRANCH=${TRAVIS_TAG}
+  echo "Using ZFS repo tag: ${ZFS_BUILD_BRANCH}"
+else
+  #Current active (master) zfs and spl development branch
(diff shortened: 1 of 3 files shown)


vishnuitta (approved):
changes are good

fetch-zfs-branch.sh
```diff
@@ -0,0 +1,49 @@
...
+  CURRENT_BRANCH=""
+  if [ -z ${TRAVIS_BRANCH} ];
+  then
+    CURRENT_BRANCH=$(git branch | grep \* | cut -d ' ' -f2)
+  else
+    CURRENT_BRANCH=${TRAVIS_BRANCH}
+  fi
+
+  #Depending on the current istgt branch name decide corresponding ZFS Build branch.
+  # If current is master - use $ZFS_MASTER
```
vishnuitta, line 23: here, comment should be 'If current is replication'

PR #82, opened 2018-09-27
fix(build): fetch the zfs header from the correct branch

kmova:
autoreconf is always fetching the zfs header file from zfs repo
master branch. This make it incompatible to build on released
branches if there has been a change to the header file on the master.

Added script to determine the zfs branch from which the header
file should be fetched.
- If tagged build, fetch from the corresponding tag
- If master (replication), fetch from the zfs master
- If non-master, fetch from corresponding zfs branch if it exists
  or use the zfs master

(cherry picked from commit c1d53eb2b414a61cb845fe55fe3e043adc429293)

### DIFF
--- a/fetch-zfs-branch.sh
+++ b/fetch-zfs-branch.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+set -e
+
+if [ ! -z ${TRAVIS_TAG} ];
+then
+  ZFS_BUILD_BRANCH=${TRAVIS_TAG}
+  echo "Using ZFS repo tag: ${ZFS_BUILD_BRANCH}"
+else
+  #Current active (master) zfs and spl development branch
+  ZFS_MASTER="zfs-0.7-release"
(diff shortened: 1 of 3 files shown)


vishnuitta (approved):
same as PR81 from replication onto v0.7.x branch

fetch-zfs-branch.sh
```diff
@@ -0,0 +1,49 @@
...
+  ZFS_BUILD_BRANCH=${TRAVIS_TAG}
+  echo "Using ZFS repo tag: ${ZFS_BUILD_BRANCH}"
+else
+  #Current active (master) zfs and spl development branch
+  ZFS_MASTER="zfs-0.7-release"
+  ZFS_REPO="https://github.com/openebs/zfs.git"
+
+  #Determinte the current branch depending on the build env.
+  CURRENT_BRANCH=""
+  if [ -z ${TRAVIS_BRANCH} ];
```
vishnuitta, line 15: ${TRAVIS_BRANCH} can never be empty.. right?

PR #83, opened 2018-09-28
feat(build): build docker container

kmova:
This repository contains the complete code for istgt
and has a dependency to dynamically fetch a zfs header
file from openebs/zfs repo.

The previous commit (#81) tried to make the fetch automated
by using the current branch name as a key to identify the
dependent branch. However, not all scenarios got covered.
Besides having the images built in openebs/cstor repo - made
it more tricky to determine the current branch that needs
to be used as key.

Also, building in opeenbs/cstor repo kind of de-links the
commits from the images and depend on a manual trigger
by the cstor repo to build a docker image.

This PR addresses the following:
- Add code to build the docker images after a successful build.
- Put a place holder for specifying the dependent zfs branch

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,6 +1,12 @@
 #!/bin/sh
 
-. ./fetch-zfs-branch.sh
+if [ ! -z ${TRAVIS_TAG} ];
+then
+  ZFS_BUILD_BRANCH=${TRAVIS_TAG}
+else
+  . ./fetch-zfs-branch.sh
+fi
+
 echo "Using zfs branch - ${ZFS_BUILD_BRANCH}"
 echo $(wget -O /tmp/zrepl_prot.h https://raw.githubusercontent.com/openebs/zfs/${ZFS_BUILD_BRANCH}/include/zrepl_prot.h)
 
--- a/docker/push
+++ b/docker/push
@@ -15,7 +15,7 @@ then
   exit 1
 fi
 
-# In case of cstor, the commit id doesn't change so using date. 
+# Generate a uniqe tag based on the commit and tag
 BUILD_ID=$(git describe --tags --always)
 
 # Determine the current branch
--- a/docker/push
+++ b/docker/push
@@ -1,0 +1,94 @@
+#!/bin/bash
+set -e
+
+if [ -z ${IMAGE_REPO} ];
+then
+  echo "Error: IMAGE_REPO is not specified";
+  exit 1
+fi
+
+IMAGEID=$( sudo docker images -q ${IMAGE_REPO}:ci )
+echo $IMAGEID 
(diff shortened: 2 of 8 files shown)


fetch-zfs-branch.sh
```diff
@@ -1,49 +1,26 @@
...
+#  fix-ta123
+#
+# Now, `fix-ta123` could be a created either from master or 
+# v0.7.x. This file needs to determine the correct zfs branch
+# by finding the parent of `fix-ta123` and checking if the 
+# correpsonding branch exists on openebs/zfs.
+#
+#For the moment, we will go with making sure the correct
+# branch name is provided as part of the release process.
+ZFS_BUILD_BRANCH="zfs-0.7-release"
```
vishnuitta, line 24: we should checkin similar file for other release branches i.e., for v0.7.x release, ZFS_BUILD_BRANCH should be v0.7.x, so that, branches created from v0.7.x will use the same one.

autogen.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/sh
 
-. ./fetch-zfs-branch.sh
+if [ ! -z ${TRAVIS_TAG} ];
+then
+  ZFS_BUILD_BRANCH=${TRAVIS_TAG}
```
vishnuitta, line 5: should we do an auto commit for fetch-zfs-branch.sh using git credentials, as part of future PRs?

docker/push
```diff
@@ -0,0 +1,94 @@
...
+
+IMAGEID=$( sudo docker images -q ${IMAGE_REPO}:ci )
+echo $IMAGEID 
+if [ -z ${IMAGEID} ];
+then
+  echo "Error: unable to get IMAGEID for ${IMAGE_REPO}:ci";
+  exit 1
+fi
+
+# In case of cstor, the commit id doesn't change so using date.
```
vishnuitta: now, by having this in istgt, commit id should be good.. right?

autogen.sh
vishnuitta: probably, this can be done along with the auto commit of tags creation during release branch creation time

fetch-zfs-branch.sh
kmova: yes, need to raise seperate PRs to update v0.7.x

autogen.sh
kmova: hmm. need to handle the case of RC tags vs actual release tag. Was thinking we should do it as part of the main release tag - branch creation stage.

docker/push
kmova: right! updated the commit. Thanks @vishnuitta

vishnuitta (approved):
changes are good

PR #94, opened 2018-10-06
fix: cstyle error in ./src/mempool_test.c

aswathkk:
fixed cstyle error in ./src/mempool_test.c
Issue: #2030


vishnuitta (approved):
changes are good

PR #133, opened 2018-10-08
[TA3500] Trigger rebuild instantly if replication factor is 1

mynktl:
Changes : 
- Trigger rebuild instantly if the `replication factor` is 1
- Removing build generated files from git
            - Makefile
            - src/build.h


vishnuitta (approved):
changes are good. But, a testcase would be required.

test_istgt.sh
```diff
@@ -527,6 +528,38 @@ run_replication_factor_test()
...
 
+	REPLICATION_FACTOR=1
+	CONSISTENCY_FACTOR=1
+	setup_test_env
+	start_replica -i "$CONTROLLER_IP" -p "$CONTROLLER_PORT" -I "$replica1_ip" -P "$replica1_port" -V $replica1_vdev &
+	replica1_pid=$!
+	sleep 2
+
+	while [ 1 ]; do
+		# We have started istgt with 20 second replica timeout
```
vishnuitta, line 539: actually, we should start istgt with normal timeout so that the condition we added will come into effect.

test_istgt.sh
```diff
@@ -527,6 +528,38 @@ run_replication_factor_test()
...
+				exit 1
+			else
+				break
+			fi
+		fi
+		sleep 10
+	done
+	kill -9 $replica1_pid
+	stop_istgt
+	rm -rf ${replica1_vdev::-1}*
```
vishnuitta, line 557: what this `::-1` does?

vishnuitta (approved):
however, merging the PR as testcase fix can be handled in other PR

test_istgt.sh
mynktl: ok.
This comment is confusing since `replica_timeout` doesn't have any relation with `replica status query` interval.
 In testing code, `replication_test` reports healthy status after receiving `replica status query` three times once rebuild starts. Due to this, the replica will take around 120 seconds to become healthy.

test_istgt.sh
mynktl: It will remove the last 1 character from the string.

PR #138, opened 2018-10-11
Remove dead code in parse_scsi_cdb()

gy741:
Hello.

I do not think that the ```inv``` variable is necessary.

Because There is no code using the ```inv``` variable.

what do you think?

Thanks.


vishnuitta (approved):
changes are good

PR #139, opened 2018-10-11
Reworked Docker image to suit style guidelines

alex3305:
The provided Docker image did not suit the Docker style guidelines as defined
by the Docker documentation [1]. This change greatly reduces the number of
layers which in effect reduces the image size by about 330MB [2]. Although this
size can differ a bit depending on the source.

This change also updates the base image to Ubuntu 16 instead of 14. I did this
because Ubuntu 14 will be end of life next year, while the Ubuntu 16 LTS will
only be EOL in 2021 [3].

 * [1] https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers
 * [2] https://imgur.com/a/SQg7ka4
 * [3] https://wiki.ubuntu.com/Releases

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,22 +2,7 @@
 # This Dockerfile builds cstor istgt container running istgt from istgt base image
 #
 
-FROM ubuntu:14.04
-RUN apt-get update; exit 0
-RUN apt-get -y install rsyslog
-RUN apt-get -y install net-tools iputils-ping
-RUN apt-get -y install libssl-dev libjson-c-dev libjemalloc-dev
-RUN apt -y install apt-file && apt-file update
-
-RUN mkdir -p /usr/local/etc/bkpistgt
-RUN mkdir -p /usr/local/etc/istgt
-COPY istgt istgtcontrol /usr/local/bin/
-COPY istgt.conf istgtcontrol.conf /usr/local/etc/bkpistgt/
-RUN touch /usr/local/etc/bkpistgt/auth.conf
-RUN touch /usr/local/etc/bkpistgt/logfile
-
-COPY entrypoint-istgtimage.sh /usr/local/bin/
-RUN chmod +x /usr/local/bin/entrypoint-istgtimage.sh
+FROM ubuntu:16.04
 
 ARG BUILD_DATE
 LABEL org.label-schema.name="cstor"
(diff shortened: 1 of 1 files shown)


utkarshmani1997 (approved):
Thanks @alex3305 for this wonderful PR, changes look good to me. Glad to see such improvements regarding size reduction in docker image. Looking for more contributions from you :).
Would you also like to review other openebs repositories for similar issues ?

alex3305:
Can this be reviewed?

@utkarshmani1997 I can look into optimizing more Dockerfile's.

utkarshmani1997:
cc : @vishnuitta @kmova @payes

docker/Dockerfile
```diff
@@ -27,5 +12,19 @@ LABEL org.label-schema.vcs-url="https://github.com/openebs/cstor"
 LABEL org.label-schema.schema-version="1.0"
 LABEL org.label-schema.build-date=$BUILD_DATE
 
-ENTRYPOINT entrypoint-istgtimage.sh
 EXPOSE 3260 6060
+
+COPY istgt istgtcontrol /usr/local/bin/
+COPY istgt.conf istgtcontrol.conf /usr/local/etc/bkpistgt/
```
vishnuitta: this directory `/usr/local/etc/bkpistgt/` will not be available by now

docker/Dockerfile
```diff
@@ -27,5 +12,19 @@ LABEL org.label-schema.vcs-url="https://github.com/openebs/cstor"
 LABEL org.label-schema.schema-version="1.0"
 LABEL org.label-schema.build-date=$BUILD_DATE
 
-ENTRYPOINT entrypoint-istgtimage.sh
 EXPOSE 3260 6060
+
```
vishnuitta, line 16: can we move this 'COPY'ing files section after apt-get install?
I'm thinking whether any package installation with 'apt-get install' may overwrite it. So, thinking that it would safe to put after all the packages installation?

docker/Dockerfile
```diff
@@ -2,22 +2,7 @@
...
-RUN mkdir -p /usr/local/etc/bkpistgt
-RUN mkdir -p /usr/local/etc/istgt
-COPY istgt istgtcontrol /usr/local/bin/
-COPY istgt.conf istgtcontrol.conf /usr/local/etc/bkpistgt/
-RUN touch /usr/local/etc/bkpistgt/auth.conf
-RUN touch /usr/local/etc/bkpistgt/logfile
-
-COPY entrypoint-istgtimage.sh /usr/local/bin/
-RUN chmod +x /usr/local/bin/entrypoint-istgtimage.sh
+FROM ubuntu:16.04
```
vishnuitta, line 5: @Moteesh did you noticed any issues by having ubuntu 16 for istgt docker image? (or) are we good to use ubuntu 16?

vishnuitta:
thanks @alex3305 for this PR.
Do we need to change the distribution that we are using in travis to build docker image?

vishnuitta:
there is good reduction in the docker image size. Below are the details with this PR:
```
REPOSITORY            TAG                 IMAGE ID            CREATED             SIZE
openebs/cstor-istgt   ci                  40284974ac24        25 seconds ago      160MB
```

docker/Dockerfile
alex3305: This directory will automatically be generated by the Docker `COPY` directive.

docker/Dockerfile
alex3305: Sure enough. Maybe it is a safer approach.

alex3305: Resolved in e7a6cadba1e8b4eb1f302e47ba982a4853e26ad7

docker/Dockerfile
vishnuitta: yes.. got to learn that from my colleague.. thanks for response @alex3305 ..

docker/Dockerfile
vishnuitta: I got info from @Moteesh that we are good to use ubuntu 16, as long as, we build the cstor-istgt image from ubuntu 16

vishnuitta (changes requested):
changes are required to change the distribution in travis that we use for building (or)
we need to create a container during build process in travis, and build the code in that container.

mittachaitu:
Closing this PR but changes are good and much required to reduce the size of the cstor-istgt image. We can rework on this with the latest changes in repo either by reopening same PR or in different PR.

PR #157, opened 2018-10-24
Refactor(istgt): modularize istgt_lu_disk_execute()

mittachaitu:
Write functions and macros for opcodes in the function istgt_lu_disk_execute.

### DIFF
--- a/src/istgt_lu_disk.c
+++ b/src/istgt_lu_disk.c
@@ -190,6 +190,117 @@ macroname:
 			##__VA_ARGS__);								\
 }
 
+#define getdata2(data, lu_cmd) {		\
+	if (lu_cmd->iobufindx == -1) {		\
+		data = NULL;					\
+		ISTGT_ERRLOG("data null\n");	\
+		lu_cmd->status = ISTGT_SCSI_STATUS_CHECK_CONDITION;\
+		return -1;							\
+	} else {							\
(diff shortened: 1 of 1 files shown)


src/istgt_lu_disk.c
```diff
@@ -190,6 +190,117 @@ extern clockid_t clockid;
...
+	if (lu_cmd->W_bit == 0) {						\
+		ISTGT_ERRLOG("c#%d %s W_bit == 0\n", conn->id, OPCODE);		\
+		lu_cmd->status = ISTGT_SCSI_STATUS_CHECK_CONDITION;		\
+		return -1;							\
+	}									\
+}
+
+#define CHECK_RSV_KEY( rsv_key )						\
+	if (rsv_key) {									\
+		rc = istgt_lu_disk_check_pr(spec, conn, PR_ALLOW(0,0,1,0,0));			\
```
payes, line 265: Values passed in PR_ALLOW varies according to the opcodes. In this PR we have taken it common for all, which is wrong. We can write a macro similar to CHECK_RSV_KEY(a,b,PR_ALLOW(c)).

payes (commented):
@mittachaitu Please update with following changes and above review comment related to reservation values.
1) Try to make single function for similar command eg(reads/writes/mode_sense) and make switch statements.
2) Make macros for invalid feild in CDB, and other similar code repetitions.

PR #158, opened 2018-10-24
[TA3365] unite tests cases for snapshot command for istgt

mynktl:
Changes : 
- Unit testing for snapshot command. Following scenarios are covered:
 _1_.  snapshot command should succeed when it is successful at least on consistency number of healthy replica
_2_. snapshot command should complete within the given time.
_3_. snapshot command failure should delete all the snapshot that is created on other replicas.
_4_. snapshot command pending on any replica should disconnect that replica if it is not complete within given time.


vishnuitta (approved):
changes are good

PR #171, opened 2018-11-01
fix(build): push prod build

vishnuitta:
This PR is to build the docker image with production (non-debug) binaries of istgt.

### DIFF
--- a/build_image.sh
+++ b/build_image.sh
@@ -1,6 +1,13 @@
 #!/bin/bash
 set -e
 
+pwd
+make clean
+sh autogen.sh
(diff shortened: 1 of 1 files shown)


kmova (approved):

build_image.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash
 set -e
 
+pwd
+make clean
+sh autogen.sh
+./configure --enable-replication
+make clean
```
mynktl, line 8: Here, `make clean` is not required.

mynktl (approved):

PR #173, opened 2018-11-03
feat(rebuild): start rebuild if there are no inflight IOs

vishnuitta:
istgt starts rebuild on a replica after 2*timeout of its start time. This is done to make sure that if any pending IOs on helping replicas are already available at the degraded replica before rebuild is started.

This PR is an optimization of starting rebuild process on the degraded replica. With this PR, rebuild will be started on a replica if all the connected replicas are not having any pending IOs with them.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -540,8 +540,10 @@ trigger_rebuild(spec_t *spec)
 	non_zero_inflight_replica_found = 0;
 	TAILQ_FOREACH(replica, &spec->rq, r_next) {
 		if (replica->replica_inflight_write_io_cnt != 0 ||
-		    replica->replica_inflight_sync_io_cnt != 0)
+		    replica->replica_inflight_sync_io_cnt != 0) {
 			non_zero_inflight_replica_found = 1;
+			break;
+		}
 	}
 
 #ifdef	DEBUG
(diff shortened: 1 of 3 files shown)


src/replication.c
```diff
@@ -528,6 +529,32 @@ trigger_rebuild(spec_t *spec)
...
+	 * are already available at degraded replica before rebuild is started.
+
+	 * optimization is that rebuild will be started on a replica if all the
+	 * connected replicas are not having any pending IOs with them.
+	 */
+	non_zero_inflight_replica_found = 0;
+	TAILQ_FOREACH(replica, &spec->rq, r_next) {
+		if (replica->replica_inflight_write_io_cnt != 0 ||
+		    replica->replica_inflight_sync_io_cnt != 0)
+			non_zero_inflight_replica_found = 1;
```
mynktl, line 544: we can add `break` here if any replica has in-flight IOs, no need to check next replica.

mynktl (approved):

src/replication.c
vishnuitta: we can't do `break`. all replicas shouldn't have any inflight IOs, as we don't know the replica which we select in rebuilding.

mynktl: yes. but once `non_zero_inflight_replica_found` is set to 1 then there is no need to check further since we have already found one replica having in-flight IOs.

vishnuitta: makes sense @mynktl .. will change that

mynktl (approved):

PR #176, opened 2018-11-10
Fixing uint64_t value in istgtcontrol's json output

mynktl:
Changes:
- Fixing negative values for `uint64_t` value in JSON format.

### DIFF
--- a/src/istgt_lu_ctl.c
+++ b/src/istgt_lu_ctl.c
@@ -56,6 +56,7 @@
 #ifdef	REPLICATION
 #include <json-c/json_object.h>
 #include "istgt_integration.h"
+#include <replication_misc.h>
 #endif
 
 #if !defined(__GNUC__)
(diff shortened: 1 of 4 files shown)


src/istgt_lu_ctl.c
```diff
@@ -3332,190 +3333,69 @@ _verb_istat ISCSIstat_rslt[ISCSI_ARYSZ] = { {0, 0, 0} };
...
+		    json_object_new_uint64(spec->totalreadblockcount));
 		json_object_object_add(jobj, "TotalWriteBlockCount",
-			jtotalwriteblockcount);
+		    json_object_new_uint64(spec->totalwriteblockcount));
 		json_object_object_add(jobj, "ReplicaCounter",
-			jreplicacount);
+		    json_object_new_int(spec->healthy_rcount));
 		json_object_object_add(jobj, "RevisionCounter",
-			jrevisioncount);
+		    json_object_new_uint64(spec->io_seq));
```
utkarshmani1997, line 3384: Please don't change the type from string to int64, JSON can only parse upto 53 bits precision, so let it be as string and others fields as well. Changing in types here require changes at other places also (jiva, exporter, maya). 
We have faced an issue earlier due to this in jiva, it was not preserving the int64 or float64 and hence the output was coming as zero (default value).

src/replication.c
```diff
@@ -3072,15 +3072,15 @@ istgt_lu_mempool_stats(char **resp)
...
-			    json_object_new_int64(
+			    json_object_new_uint64(
 			    r->replica_inflight_read_io_cnt));
 			json_object_object_add(j_replica, "in-flight write",
-			    json_object_new_int64(
+			    json_object_new_uint64(
 			    r->replica_inflight_write_io_cnt));
 			json_object_object_add(j_replica, "in-flight sync",
-			    json_object_new_int64(
+			    json_object_new_uint64(
```
utkarshmani1997, line 3083: I recommend use of string as type for the above to avoid any kind of issue at other side while parsing. Please let them (consumers of the json) know to use `json.Number` as type of the respective fields. They can easily convert it into respective types using `fieldname.Int64()` etc.

src/replication_misc.c
```diff
@@ -310,3 +312,19 @@ set_socket_keepalive(int sfd)
 out:
 	return (ret);
 }
+
+json_object *
+json_object_new_uint64(uint64_t value)
+{
+	/* 18446744073709551615 */
+	char num[21];
```
utkarshmani1997, line 320: ok so we are sending as string, ignore the comments given above, but is it certain that value will never go beyond 21 chars ?

utkarshmani1997 (changes requested):

src/replication_misc.c
vishnuitta: yes.. not beyond 21 characters

vishnuitta:
why so many changes @mynktl ? We just need to change replica guid in `istgtcontrol replica` command from int to string.

src/replication_misc.c
```diff
@@ -310,3 +312,19 @@ set_socket_keepalive(int sfd)
...
+	int r;
+	json_object *jobj;
+
+	r = snprintf(num, sizeof(num), "%" PRIu64, value);
+	if (r < 0 || (size_t)r >= sizeof(num))
+		return NULL;
```
utkarshmani1997, line 326: wrap "0" into json_object_new_string instead of returning NULL.

src/istgt_lu_ctl.c
mynktl: here, it will give output as a string.

src/replication.c
mynktl: yes. They can use parse according to JSON type. For `uint64`, we have added an API which will give json object for `uint64` as string JSON object.

src/replication_misc.c
mynktl: Here, returning `NULL` will give value as `null` (json_object_type will be `json_type_null`).

mynktl:
@vishnuitta Since there was no API for `uint64`, we have used copying value to string approach to create JSON for `uint64`. Now we have added API so I thought of using it in `istgt_lu_ctl.c` also. In `istgt_lu_ctl.c`, we were doing `malloc` for every `uint64` value and then doing free for the same.

src/replication_misc.c
utkarshmani1997: @mynktl since we are sending numeric values as string, null is invalid and may cause some errors while parsing (Since we are not using pointers at other side). So by wrapping to "0" we can avoid such scenarios.
Ref: https://stackoverflow.com/questions/10998222/json-parsing-of-int64-in-go-null-values

src/replication_misc.c
```diff
@@ -310,3 +312,19 @@ set_socket_keepalive(int sfd)
...
+json_object *
+json_object_new_uint64(uint64_t value)
+{
+	/* 18446744073709551615 */
+	char num[21];
+	int r;
+	json_object *jobj;
+
+	r = snprintf(num, sizeof(num), "%" PRIu64, value);
+	if (r < 0 || (size_t)r >= sizeof(num))
```
vishnuitta, line 325: why size_t? it should be just r >= sizeof(num)

vishnuitta (approved):
changes are good

src/replication_misc.c
mynktl: `r` is defined as `int` type. So for comparison, we are doing typecast.

src/replication_misc.c
mynktl: @utkarshmani1997 if the value is `null` then...
- It will be treated as 0 in GO if it is defined as `int/float64`.
- It will be null for a string.
I executed the same program mentioned in that link and it didn't throw any error..!

utkarshmani1997: Thanks, i was also unaware of this

PR #180, opened 2018-11-18
feat(stats): Adding vol status to istgtcontrol replica cmd

vishnuitta:
This PR is to add volume status based on various conditions to 'istgtcontrol replica' command.

With this PR, response of 'istgtcontrol -q replica vol1' will be as below:
```
{
  "Volume status": [
    {
      "name": "vol1",
      "status": "Healthy",
      "Replica status": [
        {
          "replica": "6161",
          "status": "HEALTHY",
          "checkpointed_io_seq": "1000",
          "inflight_read": "0",
          "inflight_write": "0",
          "inflight_sync": "0",
          "connected since(in seconds)": 1276
        },
        {
          "replica": "6162",
          "status": "HEALTHY",
          "checkpointed_io_seq": "1000",
          "inflight_read": "0",
          "inflight_write": "0",
          "inflight_sync": "0",
          "connected since(in seconds)": 1274
        },
        {
          "replica": "6163",
          "status": "HEALTHY",
          "checkpointed_io_seq": "1000",
          "inflight_read": "0",
          "inflight_write": "0",
          "inflight_sync": "0",
          "connected since(in seconds)": 1027
        }
      ]
    }
  ]
}
```
Below are possible 'status' values in 'volume status':
Status being Offline means no IOs are served
Status being DisabledFeatures means volume is performing in degraded mode and some features might be disabled
Status being DegradedPerformance means volume is performing in degraded mode but all features are enabled
Status being Healthy means all the components are working in healthy mode
This also have testcases to verify the volume status in 'istgtcontrol replica' command.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -1522,6 +1522,30 @@ get_cv_status(spec_t *spec, int replica_cnt, int healthy_replica_cnt)
 	return 3;
 }
 
+static const char *
+status_to_str(int status)
+{
+	const char *str;
+	switch (status) {
+		case 1:
+			str = "offline";
+			break;
+		case 2:
+			str = "disabled_feats";
+			break;
+		case 3:
+			str = "degraded";
+			break;
+		case 4:
+			str = "healthy";
+			break;
+		default:
+			str = "unknown";
+			break;
+	}
+	return str;
+}
+
 void
 istgt_lu_replica_stats(char *volname, char **resp)
 {
@@ -1559,7 +1583,7 @@ istgt_lu_replica_stats(char *volname, char **resp)
 				status = get_cv_status(spec, replica_cnt, healthy_replica_cnt);
 				MTX_UNLOCK(&spec->rq_mtx);
 				json_object_object_add(j_spec, "status",
-				    json_object_new_int64(status));
+				    json_object_new_string(status_to_str(status)));
 				json_object_object_add(j_spec,
 				    "Replica status", j_replica);
 				json_object_array_add(j_all_spec, j_spec);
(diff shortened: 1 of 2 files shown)


utkarshmani1997 (approved):
LGTM

princerachit:
@vishnuitta I will prefer the keys to have a uniform case. Some keys are space separated and can be converted to '_'(underscored) keys.
And do we really need to put `(in seconds)` in the key?
All the above suggestions are the least prioritized and can be considered as an improvement.

avishnu:
Recommend to have string states than integers as strings will be more meaningful as discussed

vishnuitta:
> Recommend to have string states than integers as strings will be more meaningful as discussed

Updated the PR description as per new changes

vishnuitta:
> @vishnuitta I will prefer the keys to have a uniform case. Some keys are space separated and can be converted to '_'(underscored) keys.
> And do we really need to put `(in seconds)` in the key?
> All the above suggestions are the least prioritized and can be considered as an improvement.

This output is for our debugging and our modules consumption. I would say - better to get them reviewed from doc team and handle in other PR if we need to

src/replication.c
```diff
@@ -1510,57 +1510,118 @@ get_replica_stats_json(replica_t *replica, struct json_object **jobj)
...
+		return 1;
+	if (healthy_replica_cnt < spec->consistency_factor)
+		return 2;
+	if (healthy_replica_cnt == spec->replication_factor)
+		return 4;
+	return 3;
+}
+
+static const char *
+status_to_str(int status)
```
mynktl, line 1526: @vishnuitta we can avoid this function and use `char *` `return value` in `get_cv_status`.

mynktl (approved):

PR #181, opened 2018-11-19
feat(istgt): Add volume status into istgtcontrol iostats command

utkarshmani1997:
Signed-off-by: Utkarsh Mani Tripathi <utkarsh.tripathi@mayadata.io>

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -1489,7 +1489,7 @@ get_replica_stats_json(replica_t *replica, struct json_object **jobj)
 
 	json_object_object_add(j_stats, "status",
 	    json_object_new_string((replica->state == ZVOL_STATUS_HEALTHY) ?
-	    "HEALTHY" : "DEGRADED"));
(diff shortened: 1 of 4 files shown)


src/replication.c
```diff
@@ -1510,40 +1510,16 @@ get_replica_stats_json(replica_t *replica, struct json_object **jobj)
 	*jobj = j_stats;
 }
 
-static int
+const char *
 get_cv_status(spec_t *spec, int replica_cnt, int healthy_replica_cnt)
 {
 	if (replica_cnt < spec->consistency_factor)
-		return 1;
+		return "Offline";
```
mynktl: It will be good if we use the macro here instead of a string. With macro in the header file, it will be easier to change values in future.

mynktl (approved):

src/replication.c
utkarshmani1997: Sounds good

mynktl (approved):

vishnuitta (approved):
changes are good

PR #182, opened 2018-11-20
refactor(cstor-volume): Make json keys camel case.

princerachit:
Some keys of output json for istgtcontrol replica command had
space separated strings or snake case. This was violating uniformity.
Made all the keys camel case and remove any description which
was present along with key name.
    eg. "Volume Status" -> "volumeStatus"

Signed-off-by: princerachit <prince.rachit@mayadata.io>


vishnuitta (approved):
changes are good

PR #183, opened 2018-11-20
refactor(cstor-volume): Change upTime type to string.

princerachit:
Changed upTime to string from int64. This is required as per
design discussion.
The upTime value would be something similar to "123454s"

Signed-off-by: princerachit <prince.rachit@mayadata.io>


vishnuitta:
`upTime` is used in test_istgt.sh script @princerachit .. so, you need change in that also for travis to pass

princerachit:
Closing this PR as this would require unnecessary work in travis scripts. upTime would continue to be used as a numeric value

PR #188, opened 2018-11-24
Error out login request if the spec is not ready

payes:
istgt allows login to happen even before required number of replicas connect to it. This causes login success, but, client will not find the disk path.
In this state, k8s tries to relogin as part of its reconcilation logic. But, login returns error saying '1 session requested, 1 already present'. And, k8s never recovers from this state.

This PR is to return error for login requests if CF number of replicas are not connected to istgt.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,7 +2,7 @@
 # This Dockerfile builds cstor istgt container running istgt from istgt base image
 #
 
-FROM ubuntu:14.04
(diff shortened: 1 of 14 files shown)


vishnuitta (approved):
changes are good

vishnuitta (approved):
changes are good

vishnuitta (approved):
changes are good

vishnuitta:
testcases need to be added

PR #189, opened 2018-11-24
fix(catchup): tgt wait for extra 5s to recv response from all replicas

vishnuitta:
Currently, istgt responds success if it receives success response from CF number of replicas. However, further IOs will be sent to all replicas. Due to this, other replicas can have many inflight IOs. This will lead to pile up of IOs on those replicas, and they may eventually get disconnected.

This PR is to make istgt wait for extra time even if CF number of replicas return success. There is also configurable option with environment variable 'extraWait'.

### DIFF
--- a/src/istgt.c
+++ b/src/istgt.c
@@ -95,6 +95,7 @@
 ISTGT g_istgt;
 #ifdef	REPLICATION
 extern int replica_timeout;
+extern int extraWait;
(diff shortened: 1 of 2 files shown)


payes (approved):
LGTM

vishnuitta:
Merging this PR as travis passed before doing 'update branch'. Also, there are issues in docker push, so, merging now as it may end up in restarting build.

PR #191, opened 2018-11-27
Checking rebuild trigger condition based in replica create time

mynktl:
Changes:
-  Checking rebuild trigger condition based in replica create time
 If there are no IOs in healthy or all other replica queues older
  than `replica->create_time` then rebuild for that replica will get
  triggered else it will be delayed.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -491,6 +491,70 @@ start_rebuild(void *buf, replica_t *replica, uint64_t data_len)
 	return ret;
 }
 
+/*
+ * Compare time..
+ * results: (T1 > T2) ? T1 : T2
+ */
+static int
+compare_time(struct timespec t1, struct timespec t2)
+{
+	if (t1.tv_sec > t2.tv_sec)
+		return 1;
+
+	if ((t1.tv_sec == t2.tv_sec) &&
+	    (t1.tv_nsec > t2.tv_nsec))
+		return 1;
+	return 0;
+}
+
+static int
+check_for_old_ios(spec_t *spec, replica_t *replica, struct timespec last)
+{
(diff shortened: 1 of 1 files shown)


src/replication.c
```diff
@@ -491,6 +491,70 @@ start_rebuild(void *buf, replica_t *replica, uint64_t data_len)
...
+	struct timespec diff1, diff2 = { 0 }, now;
+	int ret = 0;
+	replica_t *r;
+	rcmd_t *cmd;
+
+	ASSERT(MTX_LOCKED(&spec->rq_mtx));
+
+	TAILQ_FOREACH(r, &spec->rq, r_next) {
+		if ((replica && r == replica) || (!replica)) {
+			cmd = TAILQ_FIRST(&r->readyq);
```
vishnuitta, line 522: there are chances that this 'cmd' might have got deleted in replica_thread.

src/replication.c
```diff
@@ -491,6 +491,70 @@ start_rebuild(void *buf, replica_t *replica, uint64_t data_len)
...
+			if (cmd != NULL) {
+				timesdiff(CLOCK_MONOTONIC, cmd->queued_time,
+				    now, diff1);
```
vishnuitta, line 525: 'now' is not initialized

vishnuitta: ok.. timesdiff takes care of it.. but, we need to use a fixed time to know the difference

src/replication.c
mynktl: yes.

src/replication.c
mynktl: ok

PR #192, opened 2018-11-29
Adding a maximum IO wait time for replication

mynktl:
Changes:
- Checking rebuild trigger condition based on replica create time
 
   If there are no IOs queued in replication module older than `replica->create_time`
   then rebuild for that replica will get triggered else it will be delayed.

- Adding a maximum IO wait time for replication feature

  With this changes, ISTGT will wait for a fixed time until it receives
  response from all the replicas.
  If ISTGT does not receive a response from a replica within a fixed time then
  it will initiate disconnection of that replica.

  This wait time is configurable through ISTGTCONTROL.
  Command:
        1. To get the maximum IO wait time
           -- `istgtcontrol maxiowait`
        2/ To set maximum IO wait time
           -- `istgtcontrol maxiowait 3`

```
mayank@mayank:~$ sudo ./istgtcontrol -q maxiowait
5
mayank@mayank:~$ sudo ./istgtcontrol -q maxiowait 3
3
```

### DIFF
--- a/src/istgt_lu.c
+++ b/src/istgt_lu.c
@@ -76,10 +76,6 @@
 
 #define	MAX_MASKBUF 128
 
-#ifdef	REPLICATION
-extern struct timespec io_queue_time[ISTGT_MAX_NUM_LUWORKERS];
-#endif
-
 static int
 istgt_lu_allow_ipv6(const char *netmask, const char *addr)
 {
@@ -4510,7 +4506,6 @@ luworker(void *arg)
 			tdiff(second2, third, r);
 #ifdef REPLICATION
 			lu_task->lu_cmd.luworkerindx = tind;
-			clock_gettime(CLOCK_MONOTONIC, &io_queue_time[tind]);
 #endif
 			lu_task->lu_cmd.flags |= ISTGT_WORKER_PICKED;
 			rc = istgt_lu_disk_queue_start(lu, lu_num, tind);
@@ -4519,9 +4514,6 @@ luworker(void *arg)
 						lu->num, rc == -2 ? "aborted" : "failed", rc);
 			}
 
-#ifdef	REPLICATION
-			io_queue_time[tind].tv_sec = io_queue_time[tind].tv_nsec = 0;
-#endif
 			clock_gettime(clockid, &now1);
 			id = 12;
 			tdiff(third, now1, r);
--- a/src/istgt_lu.c
+++ b/src/istgt_lu.c
@@ -76,6 +76,9 @@
 
 #define	MAX_MASKBUF 128
 
+#ifdef	REPLICATION
+extern struct timespec io_queue_time[ISTGT_MAX_NUM_LUWORKERS];
+#endif
 
 static int
 istgt_lu_allow_ipv6(const char *netmask, const char *addr)
@@ -4507,6 +4510,7 @@ luworker(void *arg)
 			tdiff(second2, third, r);
 #ifdef REPLICATION
 			lu_task->lu_cmd.luworkerindx = tind;
+			clock_gettime(CLOCK_MONOTONIC, &io_queue_time[tind]);
 #endif
 			lu_task->lu_cmd.flags |= ISTGT_WORKER_PICKED;
 			rc = istgt_lu_disk_queue_start(lu, lu_num, tind);
@@ -4515,6 +4519,9 @@ luworker(void *arg)
 						lu->num, rc == -2 ? "aborted" : "failed", rc);
 			}
 
+#ifdef	REPLICATION
+			io_queue_time[tind].tv_sec = io_queue_time[tind].tv_nsec = 0;
+#endif
 			clock_gettime(clockid, &now1);
 			id = 12;
 			tdiff(third, now1, r);
--- a/src/istgt_misc.h
+++ b/src/istgt_misc.h
@@ -209,6 +209,9 @@ void poolinit(void);
 void poolfini(void);
 int poolprint(char *inbuf, int len);
 
+/*
+ * Calculate time difference
+ */
 #define timesdiff(_clockid, _st, _now, _re)				\
 {									\
 	clock_gettime(_clockid, &_now);					\
@@ -221,4 +224,17 @@ int poolprint(char *inbuf, int len);
 	}								\
 }
 
+/*
+ * Calculate difference with fixed values
+ */
+#define time_diff(_st, _end, _re)					\
+{									\
+	if ((_end.tv_nsec - _st.tv_nsec)<0) {				\
+		_re.tv_sec  = _end.tv_sec - _st.tv_sec - 1;		\
+		_re.tv_nsec = 1000000000 + _end.tv_nsec - _st.tv_nsec;	\
+	} else {							\
+		_re.tv_sec  = _end.tv_sec - _st.tv_sec;			\
+		_re.tv_nsec = _end.tv_nsec - _st.tv_nsec;		\
+	}								\
+}
 #endif /* ISTGT_MISC_H */
--- a/src/istgt_misc.h
+++ b/src/istgt_misc.h
@@ -223,18 +223,4 @@ int poolprint(char *inbuf, int len);
 		_re.tv_nsec = _now.tv_nsec - _st.tv_nsec;		\
 	}								\
 }
-
-/*
- * Calculate difference with fixed values
- */
-#define time_diff(_st, _end, _re)					\
-{									\
-	if ((_end.tv_nsec - _st.tv_nsec)<0) {				\
-		_re.tv_sec  = _end.tv_sec - _st.tv_sec - 1;		\
-		_re.tv_nsec = 1000000000 + _end.tv_nsec - _st.tv_nsec;	\
-	} else {							\
-		_re.tv_sec  = _end.tv_sec - _st.tv_sec;			\
-		_re.tv_nsec = _end.tv_nsec - _st.tv_nsec;		\
-	}								\
-}
 #endif /* ISTGT_MISC_H */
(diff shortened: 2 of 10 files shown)


src/replication.c
```diff
@@ -491,6 +492,46 @@ start_rebuild(void *buf, replica_t *replica, uint64_t data_len)
...
+	struct timespec diff1, diff2 = { 0 }, now, io_time;
+	int ret = 0;
+	int i = 0;
+
+	clock_gettime(CLOCK_MONOTONIC, &now);
+
+	for (i = 0; i < spec->luworkers; i++) {
+		io_time = io_queue_time[i];
+		if (!(io_time.tv_sec == 0 && io_time.tv_nsec == 0)) {
+			time_diff(io_time, now, diff1);
```
vishnuitta: lets have the compare_time between io_time and last instead of taking the diff

src/replication.c
```diff
@@ -491,6 +492,46 @@ start_rebuild(void *buf, replica_t *replica, uint64_t data_len)
 	return ret;
 }
 
+/*
+ * Compare time..
+ * results: (T1 > T2) ? T1 : T2
+ */
+static int
```
vishnuitta: can this be made from int to bool, which makes this a conditional operator?

src/replication.c
```diff
@@ -491,6 +492,46 @@ start_rebuild(void *buf, replica_t *replica, uint64_t data_len)
 	return ret;
 }
 
+/*
+ * Compare time..
+ * results: (T1 > T2) ? T1 : T2
```
vishnuitta: comment change from (T1 > T2) ? TRUE : FALSE

src/istgt_lu.c
```diff
@@ -4507,6 +4510,7 @@ luworker(void *arg)
 			tdiff(second2, third, r);
 #ifdef REPLICATION
 			lu_task->lu_cmd.luworkerindx = tind;
+			clock_gettime(CLOCK_MONOTONIC, &io_queue_time[tind]);
```
vishnuitta: istgt_lu_disk_queue_start does work of reading from wire also for write IOs.
So, better to move this to the 'replicate' function where IOs will be queued into replicas

src/replication.c
```diff
@@ -2679,21 +2689,26 @@ replicate(ISTGT_LU_DISK *spec, ISTGT_LU_CMD_Ptr cmd, uint64_t offset, uint64_t n
 				goto retry_read;
 			}
 
+
+			timesdiff(CLOCK_MONOTONIC, queued_time, now, diff);
 			count = 0;
-			if (success_resp == true) {
```
vishnuitta, line 2683: success_resp is not required now

src/replication.c
```diff
@@ -2679,21 +2689,26 @@ replicate(ISTGT_LU_DISK *spec, ISTGT_LU_CMD_Ptr cmd, uint64_t offset, uint64_t n
...
-				if (extra_wait.tv_sec == 0)
-					clock_gettime(CLOCK_REALTIME,
-					    &extra_wait);
-				clock_gettime(CLOCK_REALTIME, &now);
+			if (rcomm_cmd->opcode == ZVOL_OPCODE_WRITE) {
```
vishnuitta: this can be allowed for all opcodes

src/replication.c
```diff
@@ -2679,21 +2689,26 @@ replicate(ISTGT_LU_DISK *spec, ISTGT_LU_CMD_Ptr cmd, uint64_t offset, uint64_t n
...
 				for (i = 0; i < rcomm_cmd->copies_sent; i++) {
 					if (rcomm_cmd->resp_list[i].status &
 					    (RECEIVED_OK|RECEIVED_ERR))
 						count++;
+					else if (diff.tv_sec >= io_max_wait_time) {
+						ASSERT(rcomm_cmd->resp_list[i].replica);
```
vishnuitta: also, would it be better to set the rcomm_cmd->resp_list[i].status with new flag like TIMED_OUT_ERR?

vishnuitta: we need to be careful that this code doesn't executed twice for a replica, as, replica might have got deleted when this is executed first time for a replica.

src/replication.c

src/replication.c
mynktl: yes

src/replication.c

src/istgt_lu.c
mynktl: ok

src/replication.c
mynktl: yes

src/replication.c

src/replication.c

src/replication.c
```diff
@@ -24,7 +24,8 @@
 #include "istgt_scsi.h"
 #include "assert.h"
 
-int extraWait = 5;
+int io_max_wait_time = 60;
```
vishnuitta: better to make this uint64 for boundary alignment, as it will be updated in uctl thread, and, used in luworker thread.

src/replication.c
```diff
@@ -491,6 +492,42 @@ start_rebuild(void *buf, replica_t *replica, uint64_t data_len)
...
+check_for_old_ios(spec_t *spec, struct timespec last)
+{
+	struct timespec io_time;
+	int ret = 0;
+	int i = 0;
+
+	for (i = 0; i < spec->luworkers; i++) {
+		io_time = io_queue_time[i];
+		if (!(io_time.tv_sec == 0 && io_time.tv_nsec == 0)) {
+			if (!compare_time(io_time, last)) {
```
vishnuitta, line 545: adding comments would be good

src/istgt_misc.h
```diff
@@ -221,4 +224,17 @@ int poolprint(char *inbuf, int len);
 	}								\
 }
 
+/*
+ * Calculate difference with fixed values
+ */
+#define time_diff(_st, _end, _re)					\
```
vishnuitta: this is not required any more

vishnuitta (approved):
changes are good. Left few comments. Need to add testcases for istgtcontrol command, delayed IOs, partial timed out IOs, and, all timedout IOs.
@payes can you please review from the istgtcontrol command and its related code?

vishnuitta:
Need comments for newly added flag as well, and details of its usage

src/replication.c

src/replication.c
mynktl: ok

src/istgt_misc.h
mynktl: yes

payes:
Changes related to istgtcontrol look good.

src/istgtcontrol.c
```diff
@@ -977,6 +977,56 @@ exec_replica(UCTL_Ptr uctl)
...
+		rc = uctl_readline(uctl);
+		if (rc != UCTL_CMD_OK) {
+			return (rc);
+		}
+		arg = trim_string(uctl->recvbuf);
+		result = strsepq(&arg, delim);
+		strupr(result);
+		if (strcmp(result, uctl->cmd) != 0)
+			break;
+		if (uctl->iqn != NULL) {
```
vishnuitta: why this 'iqn' check?

src/replication.c
```diff
@@ -2658,9 +2691,32 @@ replicate(ISTGT_LU_DISK *spec, ISTGT_LU_CMD_Ptr cmd, uint64_t offset, uint64_t n
...
+		for (i = 0; i < rcomm_cmd->copies_sent; i++) {
+			if (rcomm_cmd->resp_list[i].status &
+			    (RECEIVED_OK|RECEIVED_ERR|REPLICATE_TIMED_OUT))
+				count++;
+			else if (diff.tv_sec >= (time_t)io_max_wait_time) {
+				ASSERT(rcomm_cmd->resp_list[i].replica);
+				rcomm_cmd->resp_list[i].status |=
+				    REPLICATE_TIMED_OUT;
+
+				MTX_LOCK(&rcomm_cmd->resp_list[i].replica->r_mtx);
```
vishnuitta: When the replica gets disconnected due to network issue, there is a chance for replica being deleted after verifying status for 'RECEIVED_ERR' or 'RECEIVED_OK'.
So, lets take spec lock, and verify that replica that we want to disconnect is still part of the list, and do inform_mgmt_conn accordingly. There may not be a need for replica's r_mtx here.

vishnuitta: add log when the replica need to be disconnected due to timeout, and also, when the replica is not part of the list.

src/replication.c
vishnuitta: success_resp variable need to be removed. Its not removed yet.

src/replication.c
```diff
@@ -590,6 +596,29 @@ trigger_rebuild(spec_t *spec)
...
+
+	if (non_zero_inflight_replica_str != NULL)
+		non_zero_inflight_replica_cnt =
+	            (unsigned int)strtol(non_zero_inflight_replica_str, NULL, 10);
+	if (non_zero_inflight_replica_cnt == 1)
+		io_queue_time[0].tv_sec = dw_replica->create_time.tv_sec - 1;
+#endif
+
+	/*
+	 * Check if healthy replica or all replica have any inflight IOs
```
vishnuitta: comment can be `check if any inflight IOs queued before dw_replica opened data connection`

src/replication.c
mynktl: ok

src/istgtcontrol.c
mynktl: This check is unnecessary. I will remove this.

src/replication.c

src/replication.c

vishnuitta (approved):
changes are good

PR #196, opened 2018-12-13
fix(iostats): correcting read bytes stats

pawanpraka1:
Signed-off-by: Pawan <pawanprakash101@gmail.com>


vishnuitta (approved):
changes are good

PR #204, opened 2018-12-29
feat(write_luworkers): set the write LUWorkers using istgtcontrol

vishnuitta:
This PR is to limit the luworkers that can do simultaneous writes in replication module.
Command to change the setting: istgtcontrol -t vol1 SET 16 <write_luworkers>

luscheduler is modified to pick only initial numbered luworkers for write IOs based on the setting.

Fixed a memory corruption in replication_test binary with read IOs during degraded mode.
Signed-off-by: Vishnu Itta vitta@mayadata.io

### DIFF
--- a/src/istgt_lu.c
+++ b/src/istgt_lu.c
@@ -4608,10 +4608,10 @@ start:
 		clock_gettime(clockid, &sch1);
 		ind = 0;
 		MTX_LOCK(&spec->schdler_mutex);
-#ifdef REPLICATION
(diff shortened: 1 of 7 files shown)


mittachaitu:
@vishnuitta Do we still need these changes? Currently, we can configure luworkers via env.

mynktl:
Closing this PR due to inactivity. Reopen this If changes/feedback required.

PR #205, opened 2018-12-31
(feat): Ephemeral adding non-quorum replicas to istgt target

mittachaitu:
This PR is to add replicas to lists (quorum and non-quorum) based on the value of quorum during management connection.

- After post rebuild on non-quorum replica, it will move from non-quorum list to quorum list.
- Write operations are allowed on the non-quorum replicas.
- `istgtcontrol replica` command will show the details of both quorum and non-quorum replica.


mittachaitu:
I will raise a PR from a different branch.

PR #206, opened 2019-01-02
(ephemeral): Allowing non-quorum replicas to connect to istgt target

mittachaitu:
This PR is to add replicas to lists (quorum and non-quorum) based on the value of quorum during management connection.

    
- After post rebuild on a non-quorum replica, it will move from non-quorum list to quorum list.
- Write operations are allowed on the non-quorum replicas.
- `istgtcontrol replica` command will show the details of both quorum and non-quorum replica.
output:
```
istgtcontrol -q replica | jq
{
  "volumeStatus": [
    {
      "name": "vol1",
      "status": "Offline",
      "replicaStatus": [
        {
          "replicaId": "6063",
          "Address": "192.168.1.14",
          "Mode": "Degraded",
          "checkpointedIOSeq": "1000",
          "inflightRead": "0",
          "inflightWrite": "0",
          "inflightSync": "0",
          "inQuorum": "1",
          "upTime": 48
        },
        {
          "replicaId": "6062",
          "Address": "192.168.1.14",
          "Mode": "Degraded",
          "checkpointedIOSeq": "1000",
          "inflightRead": "0",
          "inflightWrite": "0",
          "inflightSync": "0",
          "inQuorum": "0",
          "upTime": 28
        }
      ]
    }
  ]
}
```

### DIFF
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -264,6 +264,7 @@ handle_data_conn_error(replica_t *r)
 	int fd, data_eventfd, mgmt_eventfd2, epollfd;
 	spec_t *spec;
 	replica_t *r1 = NULL;
+	int found_in_quorum = 0;
 
 	if (r->iofd == -1) {
 		REPLICA_ERRLOG("repl %s %d %p\n", r->ip, r->port, r);
@@ -274,10 +275,20 @@ handle_data_conn_error(replica_t *r)
 	spec = r->spec;
 
 	TAILQ_FOREACH(r1, &(spec->rq), r_next)
-		if (r1 == r)
+		if (r1 == r) {
+			found_in_quorum = 1;
 			break;
+		}
+
+	if (r1 == NULL)
+		TAILQ_FOREACH(r1, &(spec->non_quorum_rq), r_next)
+			if (r1 == r) {
+				found_in_quorum = 2;
+				break;
+			}
 
 	if (r1 == NULL) {
+		MTX_LOCK(&r->r_mtx);
 		REPLICA_ERRLOG("replica %s %d not part of rqlist..\n",
 		    r->ip, r->port);
 		/*
@@ -289,16 +300,23 @@ handle_data_conn_error(replica_t *r)
 		 */
 		if (r->mgmt_eventfd2 != -1)
 			r->mgmt_eventfd2 = -1;
+		MTX_UNLOCK(&r->r_mtx);
 		MTX_UNLOCK(&spec->rq_mtx);
 		return -1;
 	}
 
-	TAILQ_REMOVE(&spec->rq, r, r_next);
-
-	if (r->state == ZVOL_STATUS_HEALTHY)
-		spec->healthy_rcount--;
-	else if (r->state == ZVOL_STATUS_DEGRADED)
-		spec->degraded_rcount--;
+	if (found_in_quorum == 1) {
+		TAILQ_REMOVE(&spec->rq, r, r_next);
+		if (r->state == ZVOL_STATUS_HEALTHY)
+			spec->healthy_rcount--;
+		else {
+			ASSERT(r->state == ZVOL_STATUS_DEGRADED);
+			spec->degraded_rcount--;
+		}
+	} else {
+		ASSERT(found_in_quorum == 2);
+		TAILQ_REMOVE(&spec->non_quorum_rq, r, r_next);
+	}
 
 	update_volstate(r->spec);
 
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -264,7 +264,7 @@ handle_data_conn_error(replica_t *r)
 	int fd, data_eventfd, mgmt_eventfd2, epollfd;
 	spec_t *spec;
 	replica_t *r1 = NULL;
-	int found_in_quorum = 0;
+	int found_in_list = 0;
 
 	if (r->iofd == -1) {
 		REPLICA_ERRLOG("repl %s %d %p\n", r->ip, r->port, r);
@@ -276,20 +276,20 @@ handle_data_conn_error(replica_t *r)
 
 	TAILQ_FOREACH(r1, &(spec->rq), r_next)
 		if (r1 == r) {
-			found_in_quorum = 1;
+			found_in_list = 1;
 			break;
 		}
 
 	if (r1 == NULL)
-		TAILQ_FOREACH(r1, &(spec->non_quorum_rq), r_next)
+		TAILQ_FOREACH(r1, &(spec->non_quorum_rq), r_non_quorum_next)
 			if (r1 == r) {
-				found_in_quorum = 2;
+				found_in_list = 2;
 				break;
 			}
 
 	if (r1 == NULL) {
 		MTX_LOCK(&r->r_mtx);
-		REPLICA_ERRLOG("replica %s %d not part of rqlist..\n",
+		REPLICA_ERRLOG("replica %s %d not part of rq and non_rq list..\n",
 		    r->ip, r->port);
 		/*
 		 * mgmt thread will check mgmt_eventfd2 fd to see if
@@ -305,7 +305,7 @@ handle_data_conn_error(replica_t *r)
 		return -1;
 	}
 
-	if (found_in_quorum == 1) {
+	if (found_in_list == 1) {
 		TAILQ_REMOVE(&spec->rq, r, r_next);
 		if (r->state == ZVOL_STATUS_HEALTHY)
 			spec->healthy_rcount--;
@@ -314,8 +314,8 @@ handle_data_conn_error(replica_t *r)
 			spec->degraded_rcount--;
 		}
 	} else {
-		ASSERT(found_in_quorum == 2);
-		TAILQ_REMOVE(&spec->non_quorum_rq, r, r_next);
+		ASSERT(found_in_list == 2);
+		TAILQ_REMOVE(&spec->non_quorum_rq, r, r_non_quorum_next);
 	}
 
 	update_volstate(r->spec);
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -276,6 +276,7 @@ handle_data_conn_error(replica_t *r)
 
 	TAILQ_FOREACH(r1, &(spec->rq), r_next)
 		if (r1 == r) {
+			assert(r->quorum == 1);
 			found_in_list = 1;
 			break;
 		}
@@ -283,6 +284,7 @@ handle_data_conn_error(replica_t *r)
 	if (r1 == NULL)
 		TAILQ_FOREACH(r1, &(spec->non_quorum_rq), r_non_quorum_next)
 			if (r1 == r) {
+				assert(r->quorum == 0);
 				found_in_list = 2;
 				break;
 			}
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -290,14 +290,14 @@ handle_data_conn_error(replica_t *r)
 			}
 
 	if (r1 == NULL) {
-		MTX_LOCK(&r->r_mtx);
 		REPLICA_ERRLOG("replica %s %d not part of rq and non_rq list..\n",
 		    r->ip, r->port);
+		MTX_LOCK(&r->r_mtx);
 		/*
 		 * mgmt thread will check mgmt_eventfd2 fd to see if
 		 * it needs to wait for replica thread or not.
 		 * At this stage, the replica hasn't been added to
-		 * spec's rqlist so we need to update mgmt_eventfd2 to -1
+		 * spec's rqlist and non_rqlist so we need to update mgmt_eventfd2 to -1
 		 * here So that mgmt_thread can skip replica thread check.
 		 */
 		if (r->mgmt_eventfd2 != -1)
--- a/src/data_conn.c
+++ b/src/data_conn.c
@@ -267,7 +267,7 @@ handle_data_conn_error(replica_t *r)
 	int found_in_list = 0;
 
 	if (r->iofd == -1) {
-		REPLICA_ERRLOG("repl %s %d %p\n", r->ip, r->port, r);
+		REPLICA_ERRLOG("repl %s %d %p data_conn error\n", r->ip, r->port, r);
 		return -1;
 	}
 
--- a/src/istgt_integration_test.c
+++ b/src/istgt_integration_test.c
@@ -392,11 +392,10 @@ handle_stats(rargs_t *rargs, zvol_io_cmd_t *zio_cmd)
 	zvol_op_stat_t *stats;
 	zvol_io_hdr_t *hdr = &(zio_cmd->hdr);
 
-	if (strcmp(zio_cmd->buf, rargs->volname) != 0)
+	if (strcmp(zio_cmd->buf, rargs->volname) != 0) {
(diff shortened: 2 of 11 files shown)


vishnuitta:
@mittachaitu travis is failing. Please look into this

mittachaitu:
> @mittachaitu travis is failing. Please look into this

@vishnuitta I am trying to fix it.

src/data_conn.c
```diff
@@ -264,21 +264,34 @@ handle_data_conn_error(replica_t *r)
...
+			if (r1 == r) {
+				assert(r->quorum == 0);
+				found_in_list = 2;
+				break;
+			}
 
 	if (r1 == NULL) {
-		REPLICA_ERRLOG("replica %s %d not part of rqlist..\n",
+		MTX_LOCK(&r->r_mtx);
+		REPLICA_ERRLOG("replica %s %d not part of rq and non_rq list..\n",
```
vishnuitta, line 293: move this print out of the lock, and below comment need to be updated by saying that 'replica hasn't been added to spec's rqlist and non_rq_list'

src/istgt_integration_test.c
```diff
@@ -393,10 +393,13 @@ handle_stats(rargs_t *rargs, zvol_io_cmd_t *zio_cmd)
 	zvol_io_hdr_t *hdr = &(zio_cmd->hdr);
 
 	if (strcmp(zio_cmd->buf, rargs->volname) != 0)
+	{
```
vishnuitta: '{' should be in the same line as 'if'

src/istgt_integration_test.c
```diff
@@ -393,10 +393,13 @@ handle_stats(rargs_t *rargs, zvol_io_cmd_t *zio_cmd)
...
+		REPLICA_ERRLOG("Vol name is not matched \n");
 		exit(1);
+	}
 
-	if (rargs->zrepl_status != ZVOL_STATUS_HEALTHY)
-		exit(1);
+//	if (rargs->zrepl_status != ZVOL_STATUS_HEALTHY)
```
vishnuitta: remove this completely

src/istgt_integration_test.c
```diff
@@ -463,6 +466,7 @@ handle_handshake(rargs_t *rargs, zvol_io_cmd_t *zio_cmd)
 	mgmt_ack->zvol_guid = rargs->replica_port;
 	mgmt_ack->port = rargs->replica_port;
 	mgmt_ack->checkpointed_io_seq = 10000;
+	mgmt_ack->quorum=1;
```
vishnuitta: follow the coding practice.. give 'space' before and after '='

src/replication.c
```diff
@@ -709,25 +733,61 @@ is_replica_newly_connected(spec_t *spec, replica_t *new_replica)
...
+ * This function checks whether given replica can be connected based on
+ * the number of already connected replicas.
+ * There is a max limit of MAXREPLICA on the number of connected replicas
+ * This functions returns true if it can be allowed, else, false
+ * This should also return false if replication_factor number of quorum replicas
+ * are connected.
+ */
 static bool
-is_rf_replicas_connected(spec_t *spec, replica_t *replica)
+can_replica_connected(spec_t *spec, replica_t *replica)
```
vishnuitta: change the name from 'can_replica_connected' to 'can_replica_connect'

src/replication.c
```diff
@@ -1154,12 +1220,27 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
 		close(iofd);
 		return -1;
 	}
+
+	if (can_replica_connected(spec, replica) == false) {
```
vishnuitta: Add a comment above saying
'After `can_replica_connect` check above, there might be any non_quorum replicas became healthy. If (healthy+degraded) reaches replication_factor, this replica addition won't be required. So, same check required here with the spec lock'

src/replication.c
```diff
@@ -1666,6 +1740,9 @@ send_replica_query(replica_t *replica, spec_t *spec, zvol_op_code_t opcode)
 	zvol_op_code_t mgmt_opcode = opcode;
 	mgmt_cmd_t *mgmt_cmd;
 
+	if ((replica->state == ZVOL_STATUS_HEALTHY) && (opcode == ZVOL_OPCODE_REPLICA_STATUS))
```
vishnuitta, line 1754: add a comment here
'This API sends the query to replica on its management connection. For healthy replicas, there is no need to send status queries'

src/replication.c
```diff
@@ -1860,20 +1917,63 @@ update_replica_status(spec_t *spec, replica_t *replica)
...
+			if (spec->healthy_rcount >= spec->replication_factor) {
+				assert(spec->degraded_rcount == 0);
+disconnect_all_non_quorum_replicas:
+				TAILQ_FOREACH(r1, &(spec->non_quorum_rq), r_non_quorum_next) {
+					REPLICA_NOTICELOG("Disconnecting non-quorum (%lu) "
+					    "as healthy: %d degraded: %d quorum replicas are available",
+					    r1->zvol_guid, spec->healthy_rcount, spec->degraded_rcount);
+					inform_mgmt_conn(r1);
+				}
+			}
```
vishnuitta, line 1987: 'update_volstate' at the end of this function causes to print a log message very often.
do update_volstate here rather than at end of function.

src/replication.c
```diff
@@ -3327,13 +3451,14 @@ cleanup_deadlist(void *arg)
...
-			for (i = 0; i < rcomm_cmd->copies_sent; i++) {
+			copies_sent = rcomm_cmd->copies_sent + rcomm_cmd->non_quorum_copies_sent;
+			for (i = 0; i < copies_sent; i++) {
 				if (rcomm_cmd->resp_list[i].status &
 				    (RECEIVED_OK|RECEIVED_ERR))
 					count++;
 			}
 
-			if (count == rcomm_cmd->copies_sent) {
+			if (count == copies_sent) {
```
vishnuitta, line 3472: similar to this, in destroy_resp_list, memory need to be checked and freed for non_quorum_copies_sent indexes also

vishnuitta: even though data_ptr will be non-null for read IOs only and non-quorum replicas will not get read IOs, its better to check and free in destroy_resp_list

src/replication_test.c
```diff
@@ -585,8 +593,8 @@ main(int argc, char **argv)
 				}
 
 				if(mgmtio->len) {
-					mgmt_data = malloc(mgmtio->len);
```
vishnuitta: it seems you overwrote recently checked in code.. need to fix this

src/data_conn.c
mittachaitu: Done sir.

src/replication.c
mittachaitu: Comment is added.

src/replication.c
mittachaitu: update_volstate is moved inside the check(if replica state is changed to healthy)

src/replication.c
mittachaitu: Modified to total no.of copies_sent instead of clearing only on replica_quorum copies_sent.

test_istgt.sh
```diff
@@ -123,35 +128,26 @@ run_and_verify_iostats() {
 		[[ $? -ne 0 ]] && echo "mount for $device_name" && exit 1
 
 		sudo dd if=/dev/urandom of=/mnt/store/file1 bs=4k count=10000 oflag=direct
-		$ISTGTCONTROL iostats
-		var1="$($ISTGTCONTROL iostats | grep -oP "(?<=TotalWriteBytes\": \")[^ ]+" | cut -d '"' -f 1)"
+		istgtcontrol iostats
```
vishnuitta: this seems to be got overwritten

test_istgt.sh
```diff
@@ -123,35 +128,26 @@ run_and_verify_iostats() {
...
 		if [ "$var2" == "$var1" ]; then
 			echo "iostats command failed, both the values are same" && exit 1
-		else
-			echo "iostats test passed"
+		else echo "iostats test passed"
 		fi
 
 		sudo umount /mnt/store
 		logout_of_volume
-		readReqTime="$($ISTGTCONTROL -q iostats | jq '.Replicas[0].ReadReqTime' | cut -d '"' -f 2)"
```
vishnuitta: here as well, it seems - some code that got checked in got removed.
Please, make sure that you hadn't removed any new code

src/replication_test.c
```diff
@@ -287,6 +291,8 @@ send_mgmt_ack(int fd, zvol_op_code_t opcode, void *buf, char *replica_ip,
 		iovec_count = 2;
 	} else {
 		build_mgmt_ack_data;
+		if(delay > 0)
```
vishnuitta: are we introducing delays in mgmt connection as well? what is the testcase that we are doing with this?
and, why doing sleep for 5 seconds? why don't we use the value in 'delay' variable?

src/replication_test.c
```diff
@@ -598,12 +611,16 @@ main(int argc, char **argv)
 					}
 				}
 				opcode = mgmtio->opcode;
-				send_mgmt_ack(mgmtfd, opcode, mgmt_data, replica_ip, replica_port, zrepl_status, &zrepl_status_msg_cnt);
+				send_mgmt_ack(mgmtfd, opcode, mgmt_data, replica_ip, replica_port, &replica_quorum, delay,zrepl_status, &zrepl_status_msg_cnt);
```
vishnuitta: instead of passing this as a parameter, make this as a global variable say replica_quorum_state and use it everywhere. Passing a local variable and changing it in other function is wrong.

src/replication_test.c
```diff
@@ -598,12 +611,16 @@ main(int argc, char **argv)
...
 			} else if (events[i].data.fd == sfd) {
 				struct sockaddr saddr;
 				socklen_t slen;
 				char hbuf[NI_MAXHOST], sbuf[NI_MAXSERV];
 				slen = sizeof(saddr);
+				if (delay > 0)
+				{
+					sleep(5);
```
vishnuitta: why this sleep as well? any test case for this?

src/replication_test.c
mittachaitu: Yes Sir, Intention is to kill non-quorum replica during mgmt connection. sleep for 5 sec is the minimum time required to error out the replica by killing. Any way I can even go with delay time. According to the test case, a small delay is enough so went with sleep for 5 sec.

src/istgt_integration_test.c
```diff
@@ -392,11 +392,10 @@ handle_stats(rargs_t *rargs, zvol_io_cmd_t *zio_cmd)
 	zvol_op_stat_t *stats;
 	zvol_io_hdr_t *hdr = &(zio_cmd->hdr);
 
-	if (strcmp(zio_cmd->buf, rargs->volname) != 0)
-		exit(1);
-
-	if (rargs->zrepl_status != ZVOL_STATUS_HEALTHY)
+	if (strcmp(zio_cmd->buf, rargs->volname) != 0) {
+		REPLICA_ERRLOG("Vol name is not matched \n");
```
vishnuitta: you better print the names that are not matching

src/replication_test.c
vishnuitta: You may need to better use different flag for introducing 'delay' in mgmt requests. Now, every mgmt request will be delayed with existing flag.

src/replication_test.c
```diff
@@ -266,8 +267,8 @@ send_mgmt_ack(int fd, zvol_op_code_t opcode, void *buf, char *replica_ip,
 			zrepl_status->state = ZVOL_STATUS_HEALTHY;
 			zrepl_status->rebuild_status = ZVOL_REBUILDING_DONE;
 			(*zrepl_status_msg_cnt) = 0;
-			if(*replica_quorum == 0)
-				state_changed_non_quorum=true;
+			if ( replica_quorum_state == 0 )
```
vishnuitta: why do we need this variable `state_changed_non_quorum`? why don't we directly set `replica_quorum_state` to 1 here?

mittachaitu: Yes, we can set the value of `replica_quorum_state` directly to 1. But IMO better to update the variable after sending the response successfully.

test_istgt.sh
```diff
@@ -518,25 +532,569 @@ run_lu_rf_test ()
...
-	start_replica -i "$CONTROLLER_IP" -p "$CONTROLLER_PORT" -I "$replica3_ip" -P "$(($replica3_port + 10))" -V $replica3_vdev &
+	pkill -9 -P $replica3_pid
+	start_replica -i "$CONTROLLER_IP" -p "$CONTROLLER_PORT" -I "$replica3_ip" -P "$(($replica3_port + 10))" -V $replica3_vdev -q "$QUORUM" &
 	replica3_pid=$!
-	sleep 5
+        sleep 5
 
-	wait $replica3_pid
-	if [ $? -eq 0 ]; then
+	if ps -p $replica3_pid > /dev/null 2>&1
```
vishnuitta, line 540: what is this?

mittachaitu: It checks whether the process exists or not.

src/replication_test.c
```diff
@@ -767,7 +770,7 @@ main(int argc, char **argv)
 						usleep(sleeptime);
 						write_ios++;
 					} else if(io_hdr->opcode == ZVOL_OPCODE_READ) {
-						if ( replica_quorum == 0)
+						if ( replica_quorum_state == 0)
 							goto error;
```
vishnuitta, line 776: put a log here

src/replication_test.c
```diff
@@ -611,15 +614,15 @@ main(int argc, char **argv)
...
-				send_mgmt_ack(mgmtfd, opcode, data, replica_ip, replica_port, &replica_quorum, delay,zrepl_status, &zrepl_status_msg_cnt);
+				send_mgmt_ack(mgmtfd, opcode, mgmt_data, replica_ip, replica_port, delay_connection, zrepl_status, &zrepl_status_msg_cnt);
 			} else if (events[i].data.fd == sfd) {
 				struct sockaddr saddr;
 				socklen_t slen;
 				char hbuf[NI_MAXHOST], sbuf[NI_MAXSERV];
 				slen = sizeof(saddr);
-				if (delay > 0)
+				if (delay_connection > 0)
 				{
```
vishnuitta: remove '{' and '}'

test_istgt.sh
```diff
@@ -209,9 +212,10 @@ write_data()
 }
 
 setup_test_env() {
+	echo "=========== $1 ============="
```
vishnuitta: why echo here?

test_istgt.sh
```diff
@@ -518,25 +530,446 @@ run_lu_rf_test ()
...
+	fi
+
+	pkill -9 -P $replica3_pid
+
+	start_replica -i "$CONTROLLER_IP" -p "$CONTROLLER_PORT" -I "$replica3_ip" -P "$replica3_port" -V $replica3_vdev &
+	replica3_pid=$!
+	date_log=$(eval date +%Y-%m-%d/%H:%M:%S.%s)
+	sleep 45
+
+	src/istgtcontrol -q replica | jq
```
vishnuitta: current macro can be used for 'istgtcontrol'

vishnuitta (approved):
changes are good. Good if we can add comments for the values selected for sleep commands in testing

PR #207, opened 2019-01-03
fix(high_cpu): issue in clock for timedwait

vishnuitta:
pthread_cond_timedwait expects absolute time at which timeout need to happen in case of no external signals. This absolute time need to be calculated using REALCLOCK clockid. Using other clockids in calculating absolute time can cause high CPU utilization or wrong timeouts.

This PR is to fix an issue in clock setting for timedwait.

This is a regression created with another PR in which latency related stats has been added.


pawanpraka1 (approved):
looks good.

PR #214, opened 2019-01-29
Adds a periodic check in check_healthy_replica_count

mittachaitu:
Periodically checking for the desired count of healthy replicas in unit test of istgt.

### DIFF
--- a/test_istgt.sh
+++ b/test_istgt.sh
@@ -565,8 +565,9 @@ run_lu_rf_test ()
 	stop_istgt
 }
 
-## check_healthy_replica_count is used to cross the count of the helathy replicas.
-check_healthy_replica_count()
+## wait_for_healthy_replicas will wait for max of 5 minutes for replicas to become healthy
+## and verifies the count of the helathy replicas with desired count.
+wait_for_healthy_replicas()
 {
 
 	no_of_replicas=$1
@@ -575,7 +576,9 @@ check_healthy_replica_count()
 	cmd="$ISTGTCONTROL -q REPLICA vol1 | jq '.\"volumeStatus\"[0].\"replicaStatus\"[].Mode' | grep -w Healthy"
 	cnt=$(eval $cmd | wc -l)
 
-	## If desired count is not met, checking for desired healthy replica count for 5 minutes with an interval of 3 seconds.
+	## Wait untill all the replicas are converted to healthy.
+	## It takes atleast 30 seconds for each replica to become healthy
+	## checking for desired healthy replica count for 5 minutes with an interval of 3 seconds.
 	if [ $cnt -ne $no_of_replicas ]
 	then
 		for (( i = 0; i < 100; i++ )) do
@@ -719,12 +722,8 @@ run_quorum_test()
 		exit 1
 	fi
 
-	## Wait untill all the replicas are converted to healthy.
-	## It takes atleast 30 seconds for each replica to become healthy
-	sleep 100
-
 	# Pass the expected healthy replica count
-	check_healthy_replica_count 3
+	wait_for_healthy_replicas 3
 
 	pkill -9 -P $replica1_pid
 	pkill -9 -P $replica2_pid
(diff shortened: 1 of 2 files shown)


mynktl (changes requested):
@mittachaitu Why are we adding 5 minutes delay for checking healthy replica count?

This wait time should be calculated according to the test scenario only. It should not be any arbitrary value.

As of now, the caller of `check_healthy_replica_count` waits for a specific duration according to the scenario and then only check for healthy replica count.

mittachaitu:
> @mittachaitu Why are we adding 5 minutes delay for checking healthy replica count?
> 
> This wait time should be calculated according to the test scenario only. It should not be any arbitrary value.
> 
> As of now, the caller of `check_healthy_replica_count` waits for a specific duration according to the scenario and then only check for healthy replica count.

@mynktl it is correct, we are waiting for the particular amount of time and checking for the healthy replica count but in some scenarios, we are not getting desired healthy replica count ( https://travis-ci.org/openebs/istgt/builds/485258499 ). Instead of an increase in wait time at caller of `check_healthy_replica_count` function, added the periodic check for healthy_replica_count for 5 minutes.

Here we are not waiting for 5 minutes if the desired healthy replica count is met within the periodic 3 seconds. If the desired count is not met (in worse case) trying for 5minutes to get the desired count at the interval of 3seconds and proceeding according to the results.

test_istgt.sh
```diff
@@ -570,14 +570,28 @@ check_healthy_replica_count()
...
 
 	no_of_replicas=$1
+	local i=0
 
 	cmd="$ISTGTCONTROL -q REPLICA vol1 | jq '.\"volumeStatus\"[0].\"replicaStatus\"[].Mode' | grep -w Healthy"
 	cnt=$(eval $cmd | wc -l)
 
-	if [ $cnt -ne $no_of_replicas ]
+	## If desired count is not met, checking for desired count for 5 minutes with an interval of 3 seconds.
+	if [ $cnt -ne $no_replicas ]
```
vishnuitta: here, is this supposed to be no_of_replicas? also, one more similar one at next few lines.

mittachaitu: Yes, it is to be `no_of_replicas` I will update it.

mynktl:
ok @mittachaitu . I understand your point. But delay before checking healthy replica should be according to test scenario only. In the test, we are also checking or testing rebuild process which ideally should take a fixed amount of time to rebuild a replica.
We can add this periodic delay, but with this, we are ensuring that max rebuild time should be 5 minutes for 3 or 2 replicas.

mittachaitu:
> ok @mittachaitu . I understand your point. But delay before checking healthy replica should be according to test scenario only. In the test, we are also checking or testing rebuild process which ideally should take a fixed amount of time to rebuild a replica.
> We can add this periodic delay, but with this, we are ensuring that max rebuild time should be 5 minutes for 3 or 2 replicas.

As of now for each replica to complete the rebuilding process in test code will take ~30 seconds. So waiting a 5 minutes is max and it's enough IMO.

mynktl:
@mittachaitu 
Ok then let's keep maximum 5 minutes wait time for the rebuilding process to complete. With this, let's remove extra wait time that is being handled in the caller of `check_healthy_replica_count` function.

test_istgt.sh
```diff
@@ -570,14 +570,28 @@ check_healthy_replica_count()
 {
```
mynktl, line 573: Please update the description of the function and also the function name.
 According to the new changes, it should be like `wait_for_healthy_replicas`.

mynktl, line 573: let's remove the wait time that is being handled in the caller of `check_healthy_replica_count`.

mittachaitu: Removed extra wait time and modified the comment and name of the function.

mynktl (approved):

vishnuitta:
@mittachaitu Travis is failing

vishnuitta (approved):
changes are good

vishnuitta:
@mittachaitu you need to cherry-pick this to v0.8.x branch as well

PR #219, opened 2019-02-11
fix(discovery): restart istgt on 4 continuous send target requests

vishnuitta:
fixes https://www.github.com/openebs/openebs/issues/2382
After successful discovery, login request will be received by target.
On continuous 4 discovery failures, high mem usage at kubelet is observed due to this issue https://github.com/open-iscsi/open-iscsi/issues/155.
This PR is to restart istgt to avoid mem consumption at kubelet.

### DIFF
--- a/src/istgt_iscsi.c
+++ b/src/istgt_iscsi.c
@@ -2059,7 +2059,7 @@ istgt_iscsi_op_login(CONN_Ptr conn, ISCSI_PDU_Ptr pdu)
 			 * As per the issue https://github.com/openebs/openebs/issues/2382,
 			 * 5 continous discovery requests failure can cause iscsi initiator
 			 * to get into high CPU usage.
-			 * This code checks for this state of 5 continuous discovery failures
+			 * This code checks for this state of 4 continuous discovery failures
 			 * and restarts the process.
 			 */
 			if (discovery_counter > 4) {
--- a/src/istgt_iscsi.c
+++ b/src/istgt_iscsi.c
@@ -5363,7 +5363,7 @@ wait_all_task(CONN_Ptr conn)
 		conn->id, conn->running_tasks);
 }
 
-
+#if 0
 static void
 snd_cleanup(void *arg)
 {
@@ -5448,6 +5448,7 @@ worker_cleanup(void *arg)
 	MTX_UNLOCK(&g_conns_mutex);
 	ISTGT_TRACELOG(ISTGT_TRACE_DEBUG, "cancel cleanup UNLOCK\n");
 }
+#endif
 
 const char lu_task_typ[4][12] = {
 	"RESPONSE",
--- a/src/istgt_iscsi.c
+++ b/src/istgt_iscsi.c
@@ -2055,6 +2055,13 @@ istgt_iscsi_op_login(CONN_Ptr conn, ISCSI_PDU_Ptr pdu)
 
 			__sync_add_and_fetch(&discovery_counter, 1);
 
+			/*
+			 * As per the issue https://github.com/openebs/openebs/issues/2382,
+			 * 5 continous discovery requests failure can cause iscsi initiator
+			 * to get into high CPU usage.
+			 * This code checks for this state of 5 continuous discovery failures
+			 * and restarts the process.
+			 */
 			if (discovery_counter > 4) {
 				ISTGT_ERRLOG("discovery counter %lu retry limit exceeded\n",
 				    discovery_counter);
--- a/src/istgt_iscsi.c
+++ b/src/istgt_iscsi.c
@@ -249,7 +249,7 @@ istgt_iscsi_read_pdu(CONN_Ptr conn, ISCSI_PDU_Ptr pdu)
 		return (-1);
 	}
 	if (rc == 0) {
-		ISTGT_TRACELOG(ISTGT_TRACE_NET, "recv() EOF (%s)\n",
+		ISTGT_ERRLOG("recv() EOF (%s)\n",
 			conn->initiator_name);
 		conn->state = CONN_STATE_EXITING;
 		return (-1);
@@ -680,8 +680,8 @@ istgt_iscsi_write_pdu_internal(CONN_Ptr conn, ISCSI_PDU_Ptr pdu, ISTGT_LU_CMD_Pt
 		rc = writev(conn->sock, &iovec[0], 5);
 		if (rc < 0) {
 			now = time(NULL);
-			ISTGT_ERRLOG("writev() failed (errno=%d,%s,time=%f) for opcode:0x%2.2x CSN:0x%x\n",
-				errno, conn->initiator_name, difftime(now, start), lu_cmd->cdb0, lu_cmd->CmdSN);
+			ISTGT_ERRLOG("writev() failed (errno=%d,%s,time=%f) for opcode:%d 0x%2.2x CSN:0x%x\n",
+				errno, conn->initiator_name, difftime(now, start), opcode, lu_cmd->cdb0, lu_cmd->CmdSN);
 			return (-1);
 		}
 		nbytes -= rc;
@@ -1744,6 +1744,8 @@ istgt_iscsi_check_values(CONN_Ptr conn)
 	return (0);
 }
 
+uint64_t discovery_counter = 0;
+
 static int
 istgt_iscsi_op_login(CONN_Ptr conn, ISCSI_PDU_Ptr pdu)
 {
@@ -1897,6 +1899,8 @@ istgt_iscsi_op_login(CONN_Ptr conn, ISCSI_PDU_Ptr pdu)
 				StatusDetail = 0x07;
 				goto response;
 			}
+			discovery_counter = 0;
+
 			snprintf(conn->target_name, sizeof (conn->target_name),
 				"%s", val);
 			snprintf(conn->target_port, sizeof (conn->target_port),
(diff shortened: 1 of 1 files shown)


payes (approved):
LGTM

src/istgt_iscsi.c
```diff
@@ -1897,6 +1899,8 @@ istgt_iscsi_op_login(CONN_Ptr conn, ISCSI_PDU_Ptr pdu)
 				StatusDetail = 0x07;
 				goto response;
 			}
+			discovery_counter = 0;
```
gila, line 1902: When 5 concurrent connections happen, this will crash even though discovery might succeed. The discovery value should be per connection and not global.

src/istgt_iscsi.c
```diff
@@ -2049,6 +2053,22 @@ istgt_iscsi_op_login(CONN_Ptr conn, ISCSI_PDU_Ptr pdu)
...
+			 * As per the issue https://github.com/openebs/openebs/issues/2382,
+			 * 5 continous discovery requests failure can cause iscsi initiator
+			 * to get into high CPU usage.
+			 * This code checks for this state of 4 continuous discovery failures
+			 * and restarts the process.
+			 */
+			if (discovery_counter > 4) {
+				ISTGT_ERRLOG("discovery counter %lu retry limit exceeded\n",
+				    discovery_counter);
+				sleep(1);
```
gila, line 2068: There is no need to sleep prior to aborting.

src/istgt_iscsi.c
```diff
@@ -2049,6 +2053,22 @@ istgt_iscsi_op_login(CONN_Ptr conn, ISCSI_PDU_Ptr pdu)
...
+			__sync_add_and_fetch(&discovery_counter, 1);
+
+			/*
+			 * As per the issue https://github.com/openebs/openebs/issues/2382,
+			 * 5 continous discovery requests failure can cause iscsi initiator
+			 * to get into high CPU usage.
+			 * This code checks for this state of 4 continuous discovery failures
+			 * and restarts the process.
+			 */
+			if (discovery_counter > 4) {
```
gila, line 2065: You might want to make this value configurable

src/istgt_iscsi.c
```diff
@@ -2049,6 +2053,22 @@ istgt_iscsi_op_login(CONN_Ptr conn, ISCSI_PDU_Ptr pdu)
...
+				ISTGT_ERRLOG("discovery counter %lu retry limit exceeded\n",
+				    discovery_counter);
+				sleep(1);
+				abort();
+			}
```
gila, line 2070: Couple of questions around this:

1. where is the core placed and what is the format? Will it be core.PID or something like that? 
2. how do we know, a core dump has happened?
3. how does one gain access to the cores or how does a user upload them?

src/istgt_iscsi.c
```diff
@@ -5343,7 +5363,7 @@ wait_all_task(CONN_Ptr conn)
 		conn->id, conn->running_tasks);
 }
 
-
+#if 0
```
gila, line 5366: These changes seem unrelated.

src/istgt_iscsi.c
```diff
@@ -5724,7 +5745,7 @@ sender(void *arg)
 	pthread_set_name_np(slf, tinfo);
 #endif
 
-	pthread_cleanup_push(snd_cleanup, (void *)conn);
+//	pthread_cleanup_push(snd_cleanup, (void *)conn);
```
gila, line 5748: This seems not needed for the change introduced as we are going to dump core anyway.

src/istgt_iscsi.c
```diff
@@ -5882,7 +5904,7 @@ sender(void *arg)
 //		MTX_UNLOCK(&conn->wpdu_mutex);
 	}
 	// MTX_UNLOCK(&conn->sender_mutex);
-	pthread_cleanup_pop(0);
+//	pthread_cleanup_pop(0);
```
gila, line 5907: Same here.

src/istgt_iscsi.c
```diff
@@ -6019,7 +6041,7 @@ worker(void *arg)
 	conn->exec_lu_task = NULL;
 	lu_task = NULL;
 
-	pthread_cleanup_push(worker_cleanup, conn);
+//	pthread_cleanup_push(worker_cleanup, conn);
```
gila, line 6044: Why is this needed?

gila (changes requested):

src/istgt_iscsi.c
vishnuitta: Every retry in discovery happens on a new connection. And, we want to track this across connections. So, it need to be global.
It can be LUN specific. But, as istgt supports only one LUN at a time, global is still fine.
I will add a comment that this need to be taken care when multiple LUN support is there in istgt.

src/istgt_iscsi.c
vishnuitta: just taking extra precaution to get the above log synced.

src/istgt_iscsi.c
vishnuitta: configurable make sense. But, it involves lot of code change for this throw away code.
An environmental variable can be added, if that makes sense. Should I go ahead with an environmental variable?

src/istgt_iscsi.c
vishnuitta: These are manual steps that need to be done by user based on the platform.

src/istgt_iscsi.c
vishnuitta: This piece of thread cleanup code is trying to take care of unexpected thread exits and is buggy. However, unexpected thread exits are not handled in the entire code, and, thus this code need to be removed.
This is related to current issue as this code uses lot of locks. These locks can impact the processing of SendTargets requests. So, removed this code in this PR.

src/istgt_iscsi.c
vishnuitta: This change is not related to this issue. But, as the other code is removed which takes care of unexpected thread exits, this code is also removed.

src/istgt_iscsi.c
vishnuitta: Explained above the intent of the code, and the reason for removing it.

src/istgt_iscsi.c
gila: I'm not sure I understand this but there is nothing that prevents -- multiple logins from different hosts in istgt right now. Keeping the counter global will trigger the if below, even though it's not what we wanted. 

Multiple nodes might issue a discovery command to this instance and if that number happens to be the whatever you set it to, it will abort() if you feel that that is not a problem, then OK. The point here is that it will cause the program to abort scenarios we do not necessarily want it to abort.

src/istgt_iscsi.c
gila: Then please call proper functions that enforce this. i.e fflush(3) seems more appropriate to obtain the desired outcome

src/istgt_iscsi.c
gila: Yeah, that might be a good idea. getenv() could be used, and default to 4 if not found. We should also make sure a log entry exists about the currently used setting.

src/istgt_iscsi.c
gila: So, the core pattern is typically something that is defined by the host. We need to understand this, or otherwise, on istgt startup make sure we archive the logs. As this problem manifests itself mostly (AFAIK) on .rpm based distros, maybe to verify what the expected pattern is would be sufficient?

On my host at least, its handled by systemd, likely its the same for those but we need to verify this. 

See, as the container will be restarted, how would we know it ever happened? Nothing I've seen either on the control plane or here would tell me that indeed this has happened.

src/istgt_iscsi.c
vishnuitta: multiple logins from different nodes will not happen at same time in k8s environment. So, it should be fine.

src/istgt_iscsi.c
vishnuitta: sure.. will do that.. thanks

src/istgt_iscsi.c
gila: Ok that makes sense. Im in favor of removing all the locks =) BUT. As this is a workaround, and we want to revert his in the near future, it will revert all. But if that's OK then I'm fine with it.

src/istgt_iscsi.c
vishnuitta: If I'm not wrong, nothing we can do on this with respect to this PR.

src/istgt_iscsi.c
gila: Thanks, so this PR addresses more things than just the discovery bug. I can't comment on the need or correctness of those issues as I do not know what they are.

src/istgt_iscsi.c
vishnuitta: similar to the above explanation

src/istgt_iscsi.c
vishnuitta: These changes are related to this issue very remotely as this code uses lot of locks and CAN cause delay in SendTargets processing if this buggy code got hit.

src/istgt_iscsi.c
vishnuitta: while reverting the changes, we need to do with care :)

vishnuitta:
This workaround is not required anymore. Issue is fixed with PR#231. Hence, closing this.

PR #230, opened 2019-02-20
fix(tcp): conn recvlowat change from 48 to 1

vishnuitta:
client connection in istgt have recvlowat set at 48. This causes istgt to wait if client sends less than 48 seconds.
During discovery process, client sends 16 bytes of data and it causes istgt to wait and finally client times out the connection.
This PR is to set the TCP recvlowat as 1, so that, istgt reads the data sent by client.

### DIFF
--- a/src/istgt_iscsi.c
+++ b/src/istgt_iscsi.c
@@ -279,11 +279,12 @@ istgt_iscsi_read_pdu(CONN_Ptr conn, ISCSI_PDU_Ptr pdu)
 	iovec[0].iov_len = 4 * pdu->total_ahs_len;
 
 	/* Header Digest */
-	iovec[1].iov_base = pdu->header_digest;
 	if (conn->header_digest) {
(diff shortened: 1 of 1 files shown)


kmova (approved):
Merging to get a debug build.

src/istgt_iscsi.c
```diff
@@ -6476,7 +6479,7 @@ istgt_create_conn(ISTGT_Ptr istgt, PORTAL_Ptr portal, int sock, struct sockaddr
 	}
 #if defined(ISTGT_USE_IOVEC)
 	/* set low water mark */
-	rc = istgt_set_recvlowat(conn->sock, ISCSI_BHS_LEN);
+	rc = istgt_set_recvlowat(conn->sock, 1);
```
gila, line 6482: Why not simply not set it to begin with?

gila:
@kmova you can simply check out the for as future reference:)

PR #231, opened 2019-02-21
fixes(istgt): leaving recvlowat as 1 to read data at earliest

vishnuitta:
This PR picks these PRs (PR#228, 229, 230) from highcpu branch, which includes,
- Leaving recvlowat as 1 to read data at earliest
- log for op_text response
- reading immediate flag from correct bit
- commented out unused code
- Not closing the connection on logout request which will be closed only when client triggers it

This PR fixes openebs/openebs#2382, openebs/openebs#2390

Adding automated tests for issue related to #2390 is pending. Adding this automation is making travis run time more than 50 minutes. So, automated tests need to be optimized to complete within 40 mins. It will be taken in another PR.

### DIFF
--- a/src/istgt_iscsi.c
+++ b/src/istgt_iscsi.c
@@ -6102,7 +6102,9 @@ worker(void *arg)
 
 		/* on socket */
 		if (events.data.fd == conn->sock) {
-			if(events.events != EPOLLIN) {
+			if ((events.events & EPOLLERR) ||
+					(events.events & EPOLLHUP) ||
+					(!(events.events & EPOLLIN))) {
 				ISTGT_ERRLOG("close conn events %d\n", events.events);
 				break;
 			}
--- a/src/istgt_iscsi.c
+++ b/src/istgt_iscsi.c
@@ -279,11 +279,12 @@ istgt_iscsi_read_pdu(CONN_Ptr conn, ISCSI_PDU_Ptr pdu)
 	iovec[0].iov_len = 4 * pdu->total_ahs_len;
 
 	/* Header Digest */
-	iovec[1].iov_base = pdu->header_digest;
 	if (conn->header_digest) {
+		iovec[1].iov_base = pdu->header_digest;
 		iovec[1].iov_len = ISCSI_DIGEST_LEN;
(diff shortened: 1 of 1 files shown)


src/istgt_iscsi.c
```diff
@@ -6019,7 +6027,7 @@ worker(void *arg)
 	conn->exec_lu_task = NULL;
 	lu_task = NULL;
 
-	pthread_cleanup_push(worker_cleanup, conn);
+//	pthread_cleanup_push(worker_cleanup, conn);
```
pawanpraka1, line 6030: do we still need this change? we should call worker_cleanup.

src/istgt_iscsi.c
```diff
@@ -6094,25 +6102,24 @@ worker(void *arg)
 
 		/* on socket */
 		if (events.data.fd == conn->sock) {
-			if ((events.events & EPOLLERR) ||
-					(events.events & EPOLLHUP) ||
-					(!(events.events & EPOLLIN))) {
-				ISTGT_ERRLOG("close conn %d\n", errno);
+			if(events.events != EPOLLIN) {
```
pawanpraka1: in tcp half close, we will get EPOLLRDHUP | EPOLLIN, we should continue in that case.

src/istgt_iscsi.c
```diff
@@ -5343,7 +5351,7 @@ wait_all_task(CONN_Ptr conn)
 		conn->id, conn->running_tasks);
 }
 
-
+#if f0
```
payes: Can we please correct this typo

src/istgt_iscsi.c
payes: we don't need it. Since this is being popped out on successful exit of worker thread, and we don't return from between.

payes (approved):

src/istgt_iscsi.c
vishnuitta: makes sense..

pawanpraka1 (approved):
looks good.

pawanpraka1:
we need to raise the PR in replication branch also.

PR #232, opened 2019-02-21
revert(workaround): remove code that restarts istgt on continuous dis…

vishnuitta:
…covery reqs

PR#228 does restart of istgt to capture core on consecutive 4 discovery requests without without a login request after the discovery request.

This PR is to revert the above behavior introduced in PR#228.


pawanpraka1 (approved):
looks good.

PR #237, opened 2019-03-07
fix(tunings): set default Luworkers in case of invalid ENV value

prateekpandey14:
1. Set default `Luworkers` value in case, if there is invalid Env value.
2. Check `queue_depth < 1` instead of 0 in case of invalid Env value

Signed-off-by: prateekpandey14 <prateek.pandey@mayadata.io>


pawanpraka1 (approved):
looks good.

vishnuitta (approved):
changes are good

vishnuitta:
@prateekpandey14 can you cherry-pick this and the first one of tunings into replication branch?
Also, same need to be done for zfs repo as well.

prateekpandey14:
Sure will do that @vishnuitta

PR #240, opened 2019-03-19
feat (Resize): support for resizing volumes in target via resize command

mittachaitu:
**What this PR does**:
 This PR supports the volume resize command in target.
 
- **Adds volume resize support in target**
Steps to resize the volume:
    1. Execute istgtcontrol resize <vol_name> \<size> command to update TARGET in-memory config and underlying volumes. On error, its returns `ERROR failed RESIZE`.
    2. Update ISTGT config file (istgt.conf) with updated size.

- **To get updated size at client side:**
     1. Execute: `sudo iscsiadm -m node -R`.
     2. Execute: `resize2fs <device_name>`

- **Adds unit test to verify volume resize**


src/replication.c
```diff
@@ -1692,39 +1742,58 @@ int istgt_lu_create_snapshot(spec_t *spec, char *snapname, int io_wait_time, int
...
 
-	if (r == false) {
-		TAILQ_FOREACH(replica, &spec->rq, r_next)
-			send_replica_snapshot(spec, replica, io_seq, snapname, ZVOL_OPCODE_SNAP_DESTROY, NULL);
+	if ((opcode == ZVOL_OPCODE_RESIZE) || (rcomm_mgmt->cmds_succeeded >= cf)) {
+		r = true;
+	}
+
+	if (r == false && opcode == ZVOL_OPCODE_SNAP_CREATE) {
+		TAILQ_FOREACH(replica, &spec->non_quorum_rq, r_non_quorum_next) {
```
mynktl: We need to send `SNAP_DESTROY` to quorum replica only.

src/replication.c
```diff
@@ -1692,39 +1742,58 @@ int istgt_lu_create_snapshot(spec_t *spec, char *snapname, int io_wait_time, int
...
+		
+		TAILQ_FOREACH(replica, &spec->non_quorum_rq, r_non_quorum_next) 
+			disconnect_nonresponding_replica(replica, io_seq, opcode);
 	}
 	spec->quiesce = 0;
 	MTX_UNLOCK(&spec->rq_mtx);
 	if (r == false)
-		REPLICA_ERRLOG("snap create ioseq: %lu resp: %d\n", io_seq, r);
-	return (ret);
+		REPLICA_ERRLOG("volume operation ioseq: %lu resp: %d\n", io_seq, r);
```
mynktl: let's log opcode also.

src/istgt_lu_ctl.c
```diff
@@ -605,6 +605,76 @@ istgt_uctl_cmd_snap(UCTL_Ptr uctl)
...
+		r = istgt_execute_volume_operation(spec, ZVOL_OPCODE_RESIZE, NULL, new_size,
+			     io_wait_time, wait_time);
+
+		if (r == true) {
+			ISTGT_LOG("Successfully resized volume %s to %s size", volname, size_str);
+			istgt_uctl_snprintf(uctl, "OK %s\n", uctl->cmd);
+			ret = UCTL_CMD_OK;
+			lu->lun[0].u.storage.size = new_size;
+			spec->size = new_size;
+			spec->blockcnt = (new_size / spec->blocklen);
```
mynktl, line 675: Let's add a comment here that .. since we are doing expansion only, we don't need to worry about queued IOs.

mynktl (changes requested):
Thanks for changes @mittachaitu .
There are few correction which needs to be addressed. Please let me know if you have any concern.

src/replication.c
mittachaitu: Done

src/replication.c
mittachaitu: Done

src/istgt_lu_ctl.c
mittachaitu: Added `NOTE:`

src/replication.c
```diff
@@ -1457,10 +1460,10 @@ disconnect_nonresponding_replica(replica_t *replica, uint64_t io_seq,
 		if (mgmt_cmd->io_hdr->io_seq == io_seq &&
 		    mgmt_cmd->io_hdr->opcode == opcode) {
 
-			if (opcode == ZVOL_OPCODE_SNAP_CREATE) {
+			if (opcode == ZVOL_OPCODE_SNAP_CREATE || opcode == ZVOL_OPCODE_RESIZE) {
```
pawanpraka1: not needed this for resize. no need to disconnect the dw replica in case it is failed on healthy helper replica as dw replica will also get resize request and it will resize. Here we just have to disconnect where this command failed.

src/istgt_lu_ctl.c
```diff
@@ -605,6 +605,77 @@ istgt_uctl_cmd_snap(UCTL_Ptr uctl)
...
+	spec = lu->lun[0].spec;
+	//old_size = lu->lun[0].u.storage.size;
+	old_size = spec->size;
+	new_size = istgt_lu_parse_size(size_str);
+	if (old_size >= new_size) {
+		ISTGT_LOG("volume(%s) size is already greater than or equal to requested resize\n", volname);
+		istgt_uctl_snprintf(uctl, "OK %s\n", uctl->cmd);
+		ret = UCTL_CMD_OK;
+	} else {
+		r = istgt_execute_volume_operation(spec, ZVOL_OPCODE_RESIZE, NULL, new_size,
```
pawanpraka1, line 665: make sure this new_size is multiple of blocklen. ZFS has strict check for this and will return error. If we d'ont verify that, zfs will end up in endless cycle of connection/disconnection.

src/replication.c
```diff
@@ -1626,53 +1669,60 @@ timeout_wait_for_command(spec_t *spec, rcommon_mgmt_cmd_t *rcomm_mgmt, int wait_
...
-	uint8_t cf = spec->consistency_factor;
-	uint8_t rf = spec->replication_factor;
+	cf = spec->consistency_factor;
+	rf = spec->replication_factor;
 
 	rcommon_mgmt_cmd_t *rmgmt = allocate_rcommon_mgmt_cmd(0);
 	replica_t *hr = spec->rebuild_info.healthy_replica;
-	if (hr) {
+	if (hr && opcode == ZVOL_OPCODE_SNAP_CREATE) {
+		BUILD_VOL_CMD_DATA(spec)
```
pawanpraka1: have this kind of macro defined as a function call.
 https://github.com/torvalds/linux/blob/ecb095bff5d4b8711a81968625b3b4a235d3e477/include/asm-generic/pgtable.h#L411

src/replication.c
```diff
@@ -1401,36 +1405,35 @@ handle_snap_prepare_resp(replica_t *replica, mgmt_cmd_t *mgmt_cmd)
...
 	mgmt_cmd->data = data;
 	mgmt_cmd->mgmt_cmd_state = WRITE_IO_SEND_HDR;
 
 	MTX_LOCK(&replica->r_mtx);
-	//TODO: Add it as the second IO, rather than tailing
-	TAILQ_INSERT_TAIL(&replica->mgmt_cmd_queue, mgmt_cmd, mgmt_cmd_next);
+
+	// Adding as second IO in mgmt_cmd queue if queue contains items
+	tmp_mgmt_cmd = TAILQ_FIRST(&replica->mgmt_cmd_queue);
+	if (tmp_mgmt_cmd != NULL) {
```
pawanpraka1, line 1432: why we need this? If I am doing resize and snapshot one by one, is it possible it will be reversed means first we will take the snapshot then resize?

src/replication.c
```diff
@@ -1626,53 +1669,60 @@ timeout_wait_for_command(spec_t *spec, rcommon_mgmt_cmd_t *rcomm_mgmt, int wait_
...
+	void *data = NULL;
+	zvol_op_resize_data_t *resize_cmd;
+
 
 	clock_gettime(CLOCK_MONOTONIC_COARSE, &last);
 	MTX_LOCK(&spec->rq_mtx);
 
-	/* Wait for any ongoing snapshot commands */
+	/* Wait for any ongoing commands which colud block IOs */
 	while (spec->quiesce == 1) {
```
pawanpraka1: why do we need to quiesce the IOs for resize?

src/replication.c
```diff
@@ -1626,53 +1669,60 @@ timeout_wait_for_command(spec_t *spec, rcommon_mgmt_cmd_t *rcomm_mgmt, int wait_
...
 
-	if (can_take_snapshot(spec) == false) {
+	if (can_send_opcode(spec, opcode) == false) {
 		MTX_UNLOCK(&spec->rq_mtx);
-		REPLICA_ERRLOG("volume is not healthy to take snapshots..\n");
+		REPLICA_ERRLOG("Volume is not healthy to process OPCODE: %d", opcode);
 		return false;
 	}
 
-	r = pause_and_timed_wait_for_ongoing_ios(spec, io_wait_time);
```
pawanpraka1, line 1660: why do we need to quiesce the IOs for resize?

src/replication.c
mittachaitu: Since we are taking lock on spec and also quiesce the other operation when other is going on (any time we will process either resize or snapshot). So there won't be any chance of having snapshots and resize in queue at given time. We are prioritizing snapshot/resize command than `ZVOL_OPCODE_REPLICA_STATUS and ZVOL_OPCODE_STATS`

src/replication.c
mittachaitu: As discussed we need to find optimal solution for resize without quiesce ongoing IOs and resize also shouldn't take precedence over earlier snapshots if it both are triggered at same time. Will be done in next PR after discussing.

src/istgt_lu_ctl.c
```diff
@@ -605,6 +605,83 @@ istgt_uctl_cmd_snap(UCTL_Ptr uctl)
...
+
+	CHECK_ARG_AND_GOTO_ERROR;
+	size_str = strsepq(&arg, delim);
+
+	CHECK_ARG_AND_GOTO_ERROR;
+	io_wait_time = atoi(strsepq(&arg, delim));
+
+	CHECK_ARG_AND_GOTO_ERROR;
+	wait_time = atoi(strsepq(&arg, delim));
+
```
pawanpraka1, line 635: Check for CHECK_ARG_AND_GOTO_ERROR here also.

pawanpraka1, line 635: Check for CHECK_ARG_AND_GOTO_ERROR here also.

pawanpraka1: as discussed this is fine.

src/replication.c
```diff
@@ -1576,13 +1588,46 @@ pause_and_timed_wait_for_ongoing_ios(spec_t *spec, int sec)
...
 }
 
+#define BUILD_VOL_CMD_DATA(_spec, _opcode, _snapname, _size, _data, _data_len)				\
+	do {												\
+		zvol_op_resize_data_t *resize_cmd;							\
+		switch (_opcode) {	 								\
+			case ZVOL_OPCODE_SNAP_CREATE:							\
+			case ZVOL_OPCODE_SNAP_DESTROY:							\
+				_data_len = strlen(_spec->volname) + strlen(_snapname) + 2;		\
+				ISTGT_LOG("IS it crashed here");					\
```
mittachaitu: TODO: Remove this log.

src/replication.c
```diff
@@ -1716,12 +1718,13 @@ int istgt_execute_volume_operation(spec_t *spec, zvol_op_code_t opcode,
...
 	cf = spec->consistency_factor;
 	rf = spec->replication_factor;
+	ret = 0;
 
 	rcommon_mgmt_cmd_t *rmgmt = allocate_rcommon_mgmt_cmd(0);
 	replica_t *hr = spec->rebuild_info.healthy_replica;
 	if (hr && opcode == ZVOL_OPCODE_SNAP_CREATE) {
-		BUILD_VOL_CMD_DATA(spec)
 		REPLICA_LOG("sending SNAP_PREP to Replica(%lu)\n", hr->zvol_guid);
+		BUILD_VOL_CMD_DATA(spec, opcode, snapname, size, data, data_len);
```
mynktl: This macro is returning a value in case of invalid opcode, without unlocking a `spec->rq_mtx`!

mynktl (changes requested):

src/replication.c
mittachaitu: I will update it.

src/replication.c
```diff
@@ -1576,13 +1588,47 @@ pause_and_timed_wait_for_ongoing_ios(spec_t *spec, int sec)
...
-		send_replica_snapshot(spec, replica, 0, snapname, ZVOL_OPCODE_SNAP_DESTROY, NULL);
-	TAILQ_FOREACH(replica, &spec->non_quorum_rq, r_non_quorum_next)
-		send_replica_snapshot(spec, replica, 0, snapname, ZVOL_OPCODE_SNAP_DESTROY, NULL);
+	zvol_op_code_t opcode = ZVOL_OPCODE_SNAP_DESTROY;
+	uint64_t size = 0;
+	size_t data_len = 0;
+	void *data = NULL;
+
+	TAILQ_FOREACH(replica, &spec->rq, r_next) {
+		BUILD_VOL_CMD_DATA(spec, opcode, snapname, size, data, data_len);
```
mynktl: @mittachaitu We are not taking a lock in this fn. Let's keep, release and return, outside of MACRO only. With that, it will also increase the readability.
Please let me know you suggestion.

mittachaitu: updated the MACRO

src/istgt_lu_ctl.c
mittachaitu: Done

mynktl (approved):

mittachaitu:
Fixed this as part of PR: https://github.com/openebs/istgt/pull/275

PR #243, opened 2019-04-07
fix(OPEN_OPCODE): send replication_factor

vishnuitta:
Fixes https://www.github.com/openebs/openebs/issues/2437

Replication factor sent from istgt during OPEN_OPCODE lets zrepl to convert the state to Healthy in case of single replica volumes.

This PR is to send replication factor in OPEN_OPCODE.
Add testcase to convert single replica to healthy


vishnuitta:
This needs PR https://github.com/openebs/cstor/pull/216 to be merged.

mynktl (approved):

PR #253, opened 2019-05-17
fix(crash): check for len in STAT opcode

vishnuitta:
istgt crashes with illegal value in ZVOL_OPCODE_STATS response.
This PR is to have strict check for len in ZVOL_OPCODE_STATS response.

This PR provides stack traces with function names during crash.


mynktl (approved):

vishnuitta:
@mynktl please review the new set of changes

mynktl (approved):
looks good to me

PR #257, opened 2019-05-24
feat(logins): allow only one login at a time.

pawanpraka1:
Whenever a new login comes from different node, logout from the old connection so that at a time only one login is present.

this is a temp fix for now, will go ahead with initiator group approach. 


vishnuitta (approved):
changes are good

vishnuitta:
automated testcases would be required for same

payes (approved):

PR #261, opened 2019-06-07
support(docker): Updating base image for istgt to latest tag{xenial-20190515}

mynktl:


vishnuitta (approved):
changes are good

PR #264, opened 2019-06-10
feat(resize): exchange volsize while creating data connection.

pawanpraka1:
Resize is done as soon as we update the istgt.conf, some replica
might be down at that time. We have to resize the internal clone
volume at that time so that IO can proceed.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -1123,7 +1123,7 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
 	memset(rio_hdr, 0, sizeof (zvol_io_hdr_t));
 	rio_hdr->opcode = ZVOL_OPCODE_OPEN;
 	rio_hdr->io_seq = 0;
-	rio_hdr->offset = 0;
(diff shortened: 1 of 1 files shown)


vishnuitta:
looks good.. but, need to be merged after merging https://github.com/openebs/libcstor/pull/17

src/replication.c
```diff
@@ -1123,7 +1123,7 @@ update_replica_entry(spec_t *spec, replica_t *replica, int iofd)
 	memset(rio_hdr, 0, sizeof (zvol_io_hdr_t));
 	rio_hdr->opcode = ZVOL_OPCODE_OPEN;
 	rio_hdr->io_seq = 0;
-	rio_hdr->offset = 0;
+	rio_hdr->volsize = spec->size;
```
vishnuitta, line 1126: we need to make sure `spec->size` is uptodate in our coming PRs for resize

vishnuitta (approved):
changes are good

PR #265, opened 2019-06-18
fix(snaprebuild) when we are transitioning to

pawanpraka1:
AFS mode and we get a snpashot request from the user,
there is a race that if snapshot command is received
after we decided to do AFS transition, we will miss or
have the inconsistant snapshot.

Here traget it self will send prepare snapshot comand to make
each replica ready to receive the snapshot command and then it will
send the command.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -313,6 +313,7 @@ check_header_sanity(zvol_io_hdr_t *resp_hdr)
 			exp_len = sizeof (zvol_op_stat_t);
 			break;
 		case ZVOL_OPCODE_SNAP_CREATE:
+		case ZVOL_OPCODE_SNAP_PREPARE:
 		case ZVOL_OPCODE_START_REBUILD:
 		case ZVOL_OPCODE_SNAP_DESTROY:
 			exp_len = 0;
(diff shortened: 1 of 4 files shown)


pawanpraka1:
dependent on https://github.com/openebs/libcstor/pull/20.

mynktl (approved):

src/replication.c
```diff
@@ -1582,18 +1596,45 @@ int istgt_lu_create_snapshot(spec_t *spec, char *snapname, int io_wait_time, int
...
+		sleep(1);
+		MTX_LOCK(&spec->rq_mtx);
+		MTX_LOCK(&rmgmt->mtx);
+		timesdiff(CLOCK_MONOTONIC_COARSE, last, now, diff);
+	}
+	MTX_UNLOCK(&rmgmt->mtx);
+
+	if (rmgmt->cmds_sent != rmgmt->cmds_succeeded) {
+		MTX_UNLOCK(&spec->rq_mtx);
+		free(rmgmt);
```
vishnuitta: this can't be deleted here as there may be pending responses from replica, and during response handling, this structure will be used.

src/replication.c
```diff
@@ -1582,18 +1596,45 @@ int istgt_lu_create_snapshot(spec_t *spec, char *snapname, int io_wait_time, int
...
+		return (false);
+	}
+
 	rcomm_mgmt = allocate_rcommon_mgmt_cmd(0);
 	free_rcomm_mgmt = 0;
 
 	r = false;
-	io_seq = ++spec->io_seq;
 	TAILQ_FOREACH(replica, &spec->rq, r_next) {
 		(void) send_replica_snapshot(spec, replica, io_seq, snapname, ZVOL_OPCODE_SNAP_CREATE, rcomm_mgmt);
```
vishnuitta, line 1685: do we need to make sure that PREPARE request is sent to all the replicas for which SNAP_CREATE is also sending? (or) its taken care by errors?

src/replication.c
```diff
@@ -2518,6 +2569,9 @@ empty_mgmt_q_of_replica(replica_t *r)
 				case ZVOL_OPCODE_SNAP_CREATE:
 					handle_snap_create_resp(r, mgmt_cmd);
 					break;
+				case ZVOL_OPCODE_SNAP_PREPARE:
+					handle_snap_prepare_resp(r, mgmt_cmd);
```
vishnuitta, line 2605: :+1:

vishnuitta (commented):
testcases required for same

src/replication.c
```diff
@@ -1557,6 +1565,43 @@ int istgt_lu_destroy_snapshot(spec_t *spec, char *snapname)
...
+/*
+ * will wait for the command till the timeout or command completetion
+ * this function should be calledwhile holding the spec->rq_mtx lock
+ */
+static inline int
+timeout_wait_for_command(spec_t *spec, rcommon_mgmt_cmd_t *rcomm_mgmt, int wait_time, struct timespec last)
+{
+	struct timespec diff, now;
+	int8_t free_rcomm_mgmt = 0;
+
```
mynktl, line 1592: Let's add assertion here to make sure that `spec->rq_mtx` is locked.

src/replication.c
mynktl: It will be handled by a replica through `is_snap_inprogress` flag.

mynktl (approved):

vishnuitta (approved):
changes are good

PR #266, opened 2019-07-08
fix(test): test case for snap prep

pawanpraka1:
adding test cases for SNAP_PREP ops.

### DIFF
--- a/src/mock_client.c
+++ b/src/mock_client.c
@@ -292,6 +292,11 @@ snapshot_thread(void *args)
 		io_wait_time = random() % 2 + 2;
 		wait_time = random() % 2 + 4;
 
+		/*
+		 * we send SNAP_PREP op code to the healthy helping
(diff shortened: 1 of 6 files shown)


mynktl (approved):

src/mock_client.c
```diff
@@ -292,9 +292,18 @@ snapshot_thread(void *args)
 		io_wait_time = random() % 2 + 2;
 		wait_time = random() % 2 + 4;
 
+		MTX_LOCK(&spec->rq_mtx);
```
vishnuitta: comment would be good to understand why we are doing this when we revisit this code

vishnuitta (approved):
changes are good.. lets handle timeout case as well in new PR

vishnuitta:
@pawanpraka1 can you look at travis failure

vishnuitta (approved):
changes are good

PR #270, opened 2019-07-23
fix(crash): graceful handling of op failure

pawanpraka1:
in case of command failure we will not get
anything in the response header data. There should
not be any assert for that, that should be handled gracefully.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -1994,8 +1994,11 @@ static void
 handle_update_spec_stats(spec_t *spec, zvol_io_hdr_t *hdr, void *resp)
 {
 	zvol_op_stat_t *stats = (zvol_op_stat_t *)resp;
-	if (hdr->status != ZVOL_OP_STATUS_OK)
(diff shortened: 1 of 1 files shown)


vishnuitta (approved):
changes are good

src/replication.c
```diff
@@ -1994,23 +1994,32 @@ static void
 handle_update_spec_stats(spec_t *spec, zvol_io_hdr_t *hdr, void *resp)
 {
 	zvol_op_stat_t *stats = (zvol_op_stat_t *)resp;
-	if (hdr->status != ZVOL_OP_STATUS_OK)
+	if ((hdr->status != ZVOL_OP_STATUS_OK) ||
+	    (hdr->len != sizeof (zvol_op_stat_t))) {
```
vishnuitta, line 1998: @pawanpraka1 this check is already part of check_header_sanity right?

pawanpraka1: yes

PR #271, opened 2019-07-25
[cherry-pick] fix(crash): graceful handling of op failure (#270)

mittachaitu:
In case of command failure we will not get
anything in the response header data. There should
not be any assert for that, that should be handled gracefully.


vishnuitta (approved):
changes are good

PR #275, opened 2019-08-20
feat (Resize): support for resizing volumes via resize command

mittachaitu:
**What this PR does**:
 This PR supports the volume resize command(with basic working flow).
 
- **Adds volume resize support in target**
Steps to resize the volume:
    1. Execute istgtcontrol resize <vol_name> \<size> command to update TARGET in-memory config and underlying volumes. On error, its returns `ERROR failed RESIZE`.
    2. Update ISTGT config file (istgt.conf) with updated size.

- **To get updated size at client side:**
     1. Execute: `sudo iscsiadm -m node -R`.
     2. Execute: `resize2fs <device_name>`

- **Adds unit test to verify volume resize**

### DIFF
--- a/src/istgt_lu_ctl.c
+++ b/src/istgt_lu_ctl.c
@@ -609,6 +609,72 @@ error_return:
 }
 
 static int
+istgt_uctl_cmd_resize(UCTL_Ptr uctl)
+{
+	ISTGT_LU_Ptr lu = NULL;
+	ISTGT_LU_DISK *spec = NULL;
+	int rc = 0, ret = UCTL_CMD_ERR;
+	char *volname, *size_str, *arg;
+	uint64_t new_size, old_size;
+	bool r;
+	const char *delim = ARGS_DELIM;
+	arg = uctl->arg;
+
+	CHECK_ARG_AND_GOTO_ERROR;
+	volname = strsepq(&arg, delim);
+
+	CHECK_ARG_AND_GOTO_ERROR;
+	size_str = strsepq(&arg, delim);
+
+	if (size_str == NULL) {
+		istgt_uctl_snprintf(uctl, "LUN format error\n");
+		goto error_return;
+	} else if (*size_str == '-') {
+		istgt_uctl_snprintf(uctl, "invalid size\n");
+		goto error_return;
+	}
+	ISTGT_LOG("volume(%s) requested to resize %s\n", volname, size_str);
+
+	lu = istgt_lu_find_target_by_volname(uctl->istgt, volname);
+	if (lu == NULL) {
+		istgt_uctl_snprintf(uctl, "ERR target not found\n");
+		goto error_return;
+	}
+	spec = lu->lun[0].spec;
+	//old_size = lu->lun[0].u.storage.size;
+	old_size = spec->size;
+	new_size = istgt_lu_parse_size(size_str);
+	if (old_size >= new_size) {
+		ISTGT_LOG("volume(%s) size is already greater than or equal to requested resize\n", volname);
+		istgt_uctl_snprintf(uctl, "OK %s\n", uctl->cmd);
+		ret = UCTL_CMD_OK;
+	} else if ((new_size % spec->blocklen) != 0) {
+		ISTGT_LOG("failed to resize volume %s new size %s "
+			"is not in multiples of block length %lu\n",
+			volname, size_str, spec->blocklen);
+		istgt_uctl_snprintf(uctl, "new size %s is not in multiples of block length %lu\n", size_str, spec->blocklen);
+		goto error_return;
+	}else {
+		r = istgt_lu_resize_volume(spec, new_size);
+
+		if (r == true) {
+			ISTGT_LOG("Successfully resized volume %s to %s size\n", volname, size_str);
+			istgt_uctl_snprintf(uctl, "OK %s\n", uctl->cmd);
+			ret = UCTL_CMD_OK;
+			lu->lun[0].u.storage.size = new_size;
+		}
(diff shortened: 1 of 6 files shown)


src/istgtcontrol.c
```diff
@@ -904,6 +904,59 @@ exec_snap(UCTL_Ptr uctl)
...
+		io_wait_time = atoi(uctl->setargv[2]);
+	else
+		io_wait_time = 10;
+
+	if (uctl->setargcnt >= 4)
+		wait_time = atoi(uctl->setargv[3]);
+	else
+		wait_time = 30;
+
+	uctl_snprintf(uctl, "%s \"%s\" \"%s\" \"%d\" \"%d\"\n",
```
mittachaitu: As of now we are not these using `io_wait_time and wait_time` should we remove it?

src/replication.c
```diff
@@ -1445,6 +1446,56 @@ send_replica_snapshot(spec_t *spec, replica_t *replica, uint64_t io_seq,
...
+	size_t data_len;
+	zvol_op_resize_data_t *resize_cmd;
+	void *data;
+	int ret = 0;
+	uint64_t num = 1;
+	mgmt_cmd_t *mgmt_cmd, *tmp_mgmt_cmd;
+
+	mgmt_cmd = malloc(sizeof(mgmt_cmd_t));
+	memset(mgmt_cmd, 0, sizeof(mgmt_cmd_t));
+	mgmt_cmd->rcomm_mgmt = rcomm_mgmt;
```
pawanpraka1: who is freeing rcomm_mgmt? there seems to be a leak here.

src/replication.c
```diff
@@ -1445,6 +1446,56 @@ send_replica_snapshot(spec_t *spec, replica_t *replica, uint64_t io_seq,
...
+	tmp_mgmt_cmd = TAILQ_FIRST(&replica->mgmt_cmd_queue);
+	if (tmp_mgmt_cmd != NULL) {
+		TAILQ_INSERT_AFTER(&replica->mgmt_cmd_queue, tmp_mgmt_cmd, mgmt_cmd, mgmt_cmd_next);
+	} else {
+		TAILQ_INSERT_TAIL(&replica->mgmt_cmd_queue, mgmt_cmd, mgmt_cmd_next);
+	}
+	MTX_UNLOCK(&replica->r_mtx);
+
+	if (rcomm_mgmt != NULL)
+		rcomm_mgmt->cmds_sent++;
```
pawanpraka1: why do we need this.

src/replication.c
```diff
@@ -1727,6 +1778,36 @@ int istgt_lu_create_snapshot(spec_t *spec, char *snapname, int io_wait_time, int
...
+	MTX_LOCK(&spec->rq_mtx);
+
+	/* Wait for any ongoing snapshot commands */
+	while (spec->quiesce == 1) {
+		MTX_UNLOCK(&spec->rq_mtx);
+		sleep(1);
+		MTX_LOCK(&spec->rq_mtx);
+	}
+	io_seq = ++spec->io_seq;
+	ISTGT_LOG("Updating io_seq: %lu as part of resize operation\n", io_seq)
```
pawanpraka1: instead of logging this, log a message which will log that resized the volume and also log the old size and new size.

src/replication.c
```diff
@@ -1727,6 +1778,36 @@ int istgt_lu_create_snapshot(spec_t *spec, char *snapname, int io_wait_time, int
...
+
+	MTX_LOCK(&spec->rq_mtx);
+
+	/* Wait for any ongoing snapshot commands */
+	while (spec->quiesce == 1) {
+		MTX_UNLOCK(&spec->rq_mtx);
+		sleep(1);
+		MTX_LOCK(&spec->rq_mtx);
+	}
+	io_seq = ++spec->io_seq;
```
pawanpraka1: remove this. not needed.

src/replication.c
mittachaitu: As of now we are passing `null` rcommon_mgmt_cmd_t so no need to release the memory.(might be extra code can we remove it?)

src/replication.c
mittachaitu: We can remove it (unnecessary code)

src/replication.c
mittachaitu: I will update log message

src/replication.c
pawanpraka1: yeah, do not even pass that argument. no need to that.

src/istgt_lu_ctl.c
```diff
@@ -608,6 +608,71 @@ istgt_uctl_cmd_snap(UCTL_Ptr uctl)
...
+		goto error_return;
+	}
+	spec = lu->lun[0].spec;
+	//old_size = lu->lun[0].u.storage.size;
+	old_size = spec->size;
+	new_size = istgt_lu_parse_size(size_str);
+	if (old_size >= new_size) {
+		ISTGT_LOG("volume(%s) size is already greater than or equal to requested resize\n", volname);
+		istgt_uctl_snprintf(uctl, "OK %s\n", uctl->cmd);
+		ret = UCTL_CMD_OK;
```
pawanpraka1: here we are returning success in this case. How the pool_mgmt  knows that it should not go ahead and update the size? Does it  verify that size is larger then only update? Better we should return error in this case.

mittachaitu: Good catch Sir I will return error

pawanpraka1 (approved):
Looks good.

vishnuitta (approved):
changes are good

PR #284, opened 2019-10-09
fix(refresh): refresh cmd handling to parse conf file for storage size

vishnuitta:
`istgt` parses istgt.conf file to read storage parameter 'capacity'.
As part of enabling 'replication', few parameters like 'file' has been removed from conf file.

New LUN creation reads the storage parameters correctly.
But, updating istgt LUN using 'istgtcontrol refresh' command has NOT been modified. It still assumes the 'file' parameter. This is causing istgt to assume wrong field as the 'capacity'.

This PR is to fix this issue by making it similar to what has been done for new LUN creation.


pawanpraka1 (approved):
Looks good.

PR #289, opened 2019-10-22
feat(replica scaledown): add support to scaledown replicas

mittachaitu:

This PR adds support to perform replica scale down (or) remove the replica from a volume without IO disruption.

**Prerequisites to perform replica scale-down:**
All replicas should be in a healthy state before triggering the replica scaledown command. 

**How to trigger scale down**
CMD: istgtcontrol drf <vol_name> <drf_value> <drf_value_no_known_replica_ids>
Ex: istgtcontrol drf vol1 2 6061 6063 (On assumption of 3 replicas 6061, 6062 and 6063 are connected and I need to remove 6062)

**TODO**
1. Move new functions into a new file(other than replication.c).

### DIFF
--- a/src/istgt_lu_ctl.c
+++ b/src/istgt_lu_ctl.c
@@ -735,6 +735,7 @@ istgt_uctl_cmd_desired_rf(UCTL_Ptr uctl)
 			    "but requested desired replication factor(%d) are "
 			    "invalid changes\n", spec->desired_replication_factor,
 			    spec->replication_factor, drf);
+			goto error_return;
 		}
 		// In case of scale down read known replica list
 		known_replica_id_list = (char **) xmalloc(sizeof(char *) * drf);
--- a/src/istgt_lu_ctl.c
+++ b/src/istgt_lu_ctl.c
@@ -741,7 +741,7 @@ istgt_uctl_cmd_desired_rf(UCTL_Ptr uctl)
 		}
 		// In case of scale down read known replica list
 		known_replica_id_list = (char **) xmalloc(sizeof(char *) * drf);
-		memset(known_replica_id_list, 0, sizeof(known_replica_id_list));
+		memset(known_replica_id_list, 0, sizeof(char *) * drf);
 		for (i=0; i < drf; i++) {
 			CHECK_ARG_AND_GOTO_ERROR;
 			known_replica_id_list[i] = strsepq(&arg, delim);
--- a/src/istgt_lu_ctl.c
+++ b/src/istgt_lu_ctl.c
@@ -690,9 +690,10 @@ istgt_uctl_cmd_desired_rf(UCTL_Ptr uctl)
 	ISTGT_LU_DISK *spec = NULL;
 	int rc = 0, ret = UCTL_CMD_ERR;
 	char *volname, *arg, *s_drf;
-	int drf;
+	int drf, i;
 	const char *delim = ARGS_DELIM;
 	arg = uctl->arg;
+	char **remaining_replicas_id;
 
 	CHECK_ARG_AND_GOTO_ERROR;
 	volname = strsepq(&arg, delim);
@@ -721,14 +722,21 @@ istgt_uctl_cmd_desired_rf(UCTL_Ptr uctl)
 	spec = lu->lun[0].spec;
 
 	if (drf < spec->desired_replication_factor) {
-		ISTGT_ERRLOG("desired replication factor(%d) is greater than"
-		    " requested desired replication factor(%d)\n",
-		    spec->desired_replication_factor, drf);
-		istgt_uctl_snprintf(uctl, "ERR current desired replication factor"
-		    " is already greater than existing desiredRF\n");
-		goto error_return;
-	}
-	if (drf == spec->desired_replication_factor) {
+		// In case of scale down read known replica list
+		remaining_replicas_id = (char **) xmalloc(sizeof(char *)*(spec->replication_factor - 1));
+		for (i=0; i < (spec->replication_factor-1); i++) {
+			remaining_replicas_id[i] = strsepq(&arg, delim);
+			ISTGT_LOG("ReplicaId %s\n", remaining_replicas_id[i]);
+		}
+		rc = istgt_lu_remove_replica(spec, remaining_replicas_id);
+
+		if (rc == 0) {
+			istgt_uctl_snprintf(uctl, "OK %s\n", uctl->cmd);
+			ret = UCTL_CMD_OK;
+		}
+		else
+			istgt_uctl_snprintf(uctl, "ERR failed %s\n", uctl->cmd);
+	} else if (drf == spec->desired_replication_factor) {
 		ISTGT_LOG("desired replication factor is already equal to "
 		    "requested desired replication factor\n");
 		istgt_uctl_snprintf(uctl, "OK %s\n", uctl->cmd);
--- a/src/istgt_lu_ctl.c
+++ b/src/istgt_lu_ctl.c
@@ -744,7 +744,6 @@ istgt_uctl_cmd_desired_rf(UCTL_Ptr uctl)
 			known_replica_id_list[i] = strsepq(&arg, delim);
 		}
 		rc = istgt_lu_remove_unknown_replica(spec, drf, known_replica_id_list);
-		rc = 0;
 		if (rc == 0) {
 			istgt_uctl_snprintf(uctl, "OK %s\n", uctl->cmd);
 			ret = UCTL_CMD_OK;
--- a/src/istgt_lu_ctl.c
+++ b/src/istgt_lu_ctl.c
@@ -693,7 +693,7 @@ istgt_uctl_cmd_desired_rf(UCTL_Ptr uctl)
 	int drf, i;
 	const char *delim = ARGS_DELIM;
 	arg = uctl->arg;
-	char **remaining_replicas_id;
+	char **known_replica_id_list = NULL;
 
 	CHECK_ARG_AND_GOTO_ERROR;
 	volname = strsepq(&arg, delim);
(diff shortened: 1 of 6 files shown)


src/replication.c
```diff
@@ -3607,6 +3794,8 @@ replicate(ISTGT_LU_DISK *spec, ISTGT_LU_CMD_Ptr cmd, uint64_t offset, uint64_t n
 	num_read_ios = 0;
 
 	TAILQ_FOREACH(replica, &spec->rq, r_next) {
+		if (replica->cordon == 1)
```
mittachaitu, line 3813: Need to add the same check while sending IO to non-quorum replicas

src/istgt_lu_ctl.c
```diff
@@ -721,14 +722,38 @@ istgt_uctl_cmd_desired_rf(UCTL_Ptr uctl)
...
+			istgt_uctl_snprintf(uctl, "ERR current desired "
+			    "replication factor(%d) and replication factor(%d) "
+			    "but requested desired replication factor(%d) are "
+			    "invalid changes\n", spec->desired_replication_factor,
+			    spec->replication_factor, drf);
+			goto error_return;
+		}
+		// In case of scale down read known replica list
+		known_replica_id_list = (char **) xmalloc(sizeof(char *) * drf);
+		memset(known_replica_id_list, 0, sizeof(known_replica_id_list));
```
mynktl: It should be `memset(,,sizeof(char *) * drf)`

src/istgtcontrol.c
```diff
@@ -912,9 +914,31 @@ exec_command(UCTL_Ptr uctl)
...
 	char *name = uctl->setargv[0];
 	char *value = uctl->setargv[1];
-
-	uctl_snprintf(uctl, "%s \"%s\" \"%s\" \n",
-		uctl->cmd, name, value);
+	int i = 0, listcnt;
+	char *id_list = NULL;
+
+	listcnt = uctl->setargcnt - 2;
+	if (listcnt != 0) {
```
mynktl: This should be `if (listcnt > 0)`.

src/istgtcontrol.c
```diff
@@ -912,9 +914,31 @@ exec_command(UCTL_Ptr uctl)
...
+		// Since list of replicaIds sent via socket
+		// each replicaId should enclosed with in "" and space
+		id_list = (char *) malloc(sizeof(char) * ((listcnt * 
+		    REPLICA_ID_LEN) + (listcnt * 4)));
+		memset(id_list, 0, sizeof(id_list));
```
mynktl: wrong size `sizeof(id_list)`.

src/istgtcontrol.c
```diff
@@ -912,9 +914,31 @@ exec_command(UCTL_Ptr uctl)
...
+		for (i=0; i < listcnt; i++) {
+			strcat(id_list, "\"");
+			strncat(id_list, uctl->setargv[i+2], REPLICA_ID_LEN);
+			strcat(id_list, "\" ");
+		}
+	}
+	//Release the momory
+	if (listcnt != 0) {
```
mynktl: This should be `if (listcnt > 0)`.

src/replication.c
```diff
@@ -2173,6 +2235,211 @@ istgt_lu_resize_volume(spec_t *spec, uint64_t size) {
...
+#define IS_REPLICA_ID_EXIST_IN_LIST(_replica_id, _list, _no_of_items, _is_exist) 	\
+	do {										\
+		int i;									\
+		for (i=0; i < _no_of_items; i++) {					\
+			if (strncmp(_replica_id, _list[i], REPLICA_ID_LEN) == 0) {	\
+					_is_exist = true;				\
+					break;						\
+			}								\
+		}									\
+	}while(0)
```
mynktl: missing space.. `} while (0)`.

src/replication.c
```diff
@@ -2173,6 +2235,211 @@ istgt_lu_resize_volume(spec_t *spec, uint64_t size) {
...
+
+
+#ifdef REPLICATION
```
mynktl: Not needed. Code from this file will get compiled if `REPLICATION` is defined.

src/replication.c
```diff
@@ -2173,6 +2235,211 @@ istgt_lu_resize_volume(spec_t *spec, uint64_t size) {
...
+	replica_t *replica, *removing_replica = NULL;
+	int rc = -1;
+	// Hard coded IO wait time
+	int io_wait_time = 5;
+	int new_rf = drf;
+	int new_cf = CONSISTENCY_FACTOR(new_rf);
+	int found_in_rq_list = false;
+	int found_replica = false;
+
+#ifdef REPLICATION
```
mynktl: not needed.

src/replication.c
```diff
@@ -2173,6 +2235,211 @@ istgt_lu_resize_volume(spec_t *spec, uint64_t size) {
...
+	trusty_replica_t *trusty_replica = NULL;
+#endif
+
+	MTX_LOCK(&spec->rq_mtx);
+	assert(drf == spec->replication_factor - 1);
```
mynktl: let's use wrapper`ASSERT` for `assert` here.

src/replication.c
```diff
@@ -2173,6 +2235,211 @@ istgt_lu_resize_volume(spec_t *spec, uint64_t size) {
...
+
+	if (spec->healthy_rcount < spec->consistency_factor)
+		return false;
+
+	// TODO: Return true when all the replicas_id should be healthy
+	// else return false
+	TAILQ_FOREACH(replica, &spec->rq, r_next) {
+		if (replica != removing_replica &&
+		    replica->state != ZVOL_STATUS_HEALTHY)
+			return false;
```
mynktl, line 2233: Can't we remove the replica if `healthy_rcount > (spec->cf + 1)` instead of checking all the replica's state?

src/replication.c
```diff
@@ -2173,6 +2235,211 @@ istgt_lu_resize_volume(spec_t *spec, uint64_t size) {
...
+
+update_volume_configurations:
+	spec->replication_factor = new_rf;
+	spec->consistency_factor = new_cf;
+	spec->lu->replication_factor = new_rf;
+	spec->lu->consistency_factor = new_cf;
+	spec->lu->desired_replication_factor = drf;
+	spec->desired_replication_factor = drf;
+
+#ifdef REPLICATION
```
mynktl: not needed

src/replication.c
```diff
@@ -2173,6 +2235,211 @@ istgt_lu_resize_volume(spec_t *spec, uint64_t size) {
...
+			TAILQ_REMOVE(&spec->lu->trusty_replicas,
+			    trusty_replica, next);
+			break;
+		}
+	}
+#endif
+	spec->quiesce = 0;
+
+	if (removing_replica != NULL)
+		inform_mgmt_conn(removing_replica);
```
mynktl, line 2369: `inform_mgmt_conn` updates the replica parameter, so better to take a replica lock prior to this function call.

mynktl (changes requested):

src/istgtcontrol.c
mittachaitu: Definitely It will not be < 0 but anyhow to be specific I will update with `listcnt > 0`

src/istgt_lu_ctl.c
mittachaitu: I will take care of memory things somehow I was in miss assumption

src/replication.c
mittachaitu: I think if we go this approach then we will end up having different conditions based on `replication factor` and one more gotcha with this is we can't perform replica scale down if `spec->rf is 2`.

src/istgtcontrol.c
mittachaitu: Handled

src/istgtcontrol.c
mittachaitu: handled

src/istgt_lu_ctl.c
mittachaitu: handled

src/replication.c
mittachaitu: As discussed I will update to `healthy_rcount >= (spec->cf + 1)`

src/replication.c
mittachaitu: Done

src/replication.c
mittachaitu: Removed

src/replication.c
mittachaitu: updated

src/istgtcontrol.c
```diff
@@ -84,6 +84,7 @@
 
 #define	MAX_LINEBUF 4096
 #define	UCTL_CHAP_CHALLENGE_LEN 1024
+#define REPLICA_ID_LEN 32
```
mynktl: `REPLICA_ID_LEN` should be consistent with libcStor. Instead of defining it here, let's use it's definition from `zrepl_prot.h` only.

src/replication.c
mynktl: ok. That make sense. Let's mention this condition, for replica scale-down, in PR description.

src/replication.c
```diff
@@ -2173,6 +2175,208 @@ istgt_lu_resize_volume(spec_t *spec, uint64_t size) {
...
+	int new_cf = CONSISTENCY_FACTOR(new_rf);
+	int found_in_rq_list = false;
+	int found_replica = false;
+
+	trusty_replica_t *trusty_replica = NULL;
+
+	MTX_LOCK(&spec->rq_mtx);
+	ASSERT(drf == spec->replication_factor - 1);
+
+#ifdef REPLICATION
```
mynktl: Please remove this macro protection.

mynktl (commented):
@mittachaitu Few additional changes are required. `replication.c` is completely meant for replication functionality only, no need to use `#ifdef REPLICATION` in this file.

src/replication.c
mittachaitu: Sorry for not going through changes I will update it

src/istgtcontrol.c
mittachaitu: Updated

src/replication.c
mittachaitu: Updated as prerequisites step to perform scaledown step

mynktl (approved):

payes (approved):

PR #291, opened 2019-11-04
fix(unit test): fix single replica rebuild test case

mittachaitu:
This PR is to fix single replica rebuild test case. This commit fixes the flaky test cases failure in travis..

### DIFF
--- a/src/replication_test.c
+++ b/src/replication_test.c
@@ -65,6 +65,7 @@ uint64_t read_ios;
 uint64_t write_ios;
 int replica_quorum_state = 0;
 char replica_id[REPLICA_ID_LEN];
+uint64_t replication_factor = 0;
 
 static void
 sig_handler(int sig)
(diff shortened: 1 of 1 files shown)


src/replication_test.c
```diff
@@ -249,6 +250,17 @@ test_read_data(int fd, uint8_t *data, uint64_t len)
 	return nbytes;
 }
 
+static bool
+is_valid_rebuild_status(zrepl_status_ack_t *zrepl_status) {
+	if (zrepl_status->state == ZVOL_STATUS_DEGRADED)
+		return true;
+	if (replication_factor == 1 &&
```
pawanpraka1, line 257: do we have a test case for this scenario in istgt?

mittachaitu: I think there are no test cases for this scenario. Since at target side, rebuilding is triggered only after the response from opcode `ZVOL_OPCODE_REPLICA_STATUS`. In this, if we get a `Healthy` status rebuild itself will not trigger.

mynktl (approved):

pawanpraka1 (approved):
looks good.

PR #301, opened 2020-01-10
fix(cleanup,mempool): Cleanup mempool related thread while destroying a volume

mynktl:

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -4380,13 +4380,14 @@ destroy_rcommon_deadlist(spec_t *spec)
     int mempool_stale_entry, i;
     rcommon_cmd_t *rcomm_cmd;
 
-    mempool_stale_entry= get_num_entries_from_mempool(&spec->rcommon_deadlist);
+    mempool_stale_entry = get_num_entries_from_mempool(&spec->rcommon_deadlist);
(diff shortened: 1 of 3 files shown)


src/replication.c
```diff
@@ -4352,7 +4354,10 @@ initialize_replication()
 void
 destroy_volume(spec_t *spec)
 {
-	ASSERT0(get_num_entries_from_mempool(&spec->rcommon_deadlist));
+    destroy_rcommon_deadlist(spec);
+
+    pthread_cancel(spec->deadlist_cleanup_thread);
```
vishnuitta, line 4362: should we set cancel_type as asynchronous? it seems - default is deferred and it needs cancellation point

vishnuitta:
``` A thread's  cancellation type, determined by pthread_setcanceltype(3), may be either asynchronous or deferred (the default for new threads).  Asynchronous cancelability means that the thread
       can be canceled at any time (usually immediately, but the system does not guarantee this).  Deferred cancelability means that cancellation will be delayed until the thread next calls a  func‐
       tion that is a cancellation point.  A list of functions that are or may be cancellation points is provided in pthreads(7).
```

vishnuitta (approved):
changes are good

PR #303, opened 2020-01-10
fix(crash): cleanup mempool related thread while destroying volume (301)

vishnuitta:
cherry pick https://github.com/openebs/istgt/pull/301

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -4354,19 +4354,19 @@ initialize_replication()
 void
 destroy_volume(spec_t *spec)
 {
-    int ret = 0;
-    void *res;
+	int ret = 0;
+	void *res;
 
-    destroy_rcommon_deadlist(spec);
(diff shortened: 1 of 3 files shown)


src/replication.c
```diff
@@ -4352,7 +4354,19 @@ initialize_replication()
 void
 destroy_volume(spec_t *spec)
 {
-	ASSERT0(get_num_entries_from_mempool(&spec->rcommon_deadlist));
+    int ret = 0;
+    void *res;
+
+    destroy_rcommon_deadlist(spec);
```
pawanpraka1: shouldn't we clean rcommon_deadlist after cancelling the thread. It's possible that thread is working on a pointer which is just freed by this destroy?

vishnuitta: thats a good one @pawanpraka1

mynktl (approved):

PR #304, opened 2020-01-10
fix(crash): cleanup mempool related thread while destroying volume

mynktl:


mynktl:
Fix is merged by  #303.

PR #310, opened 2020-02-19
feat(build): enable auto build for arm64 binaries

shubham14bajpai:
Signed-off-by: shubham <shubham.bajpai@mayadata.io>

- auto build and push arm64 and amd64 binaries

### DIFF
--- a/build_image.sh
+++ b/build_image.sh
@@ -14,7 +14,7 @@ if [ "${ARCH}" = "x86_64" ]; then
 elif [ "${ARCH}" = "aarch64" ]; then
 	./configure --enable-replication --build=arm-linux
 else
-	echo "Unusable architecture: ${ARCH}"
+	echo "Unsupported architecture: ${ARCH}"
 	exit 1
 fi
 make clean
--- a/build_image.sh
+++ b/build_image.sh
@@ -8,7 +8,15 @@ ARCH=$(uname -m)
 
 make clean
 bash autogen.sh
-./configure --enable-replication
+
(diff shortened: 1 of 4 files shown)


build_image.sh
```diff
@@ -8,7 +8,15 @@ ARCH=$(uname -m)
...
 make clean
 bash autogen.sh
-./configure --enable-replication
+
+if [ "${ARCH}" = "x86_64" ]; then
+	./configure --enable-replication
+elif [ "${ARCH}" = "aarch64" ]; then
+	./configure --enable-replication --build=arm-linux
+else
+	echo "Unusable architecture: ${ARCH}"
```
kmova: n_: unusable -> unsupported

src/arch/arm/rte_atomic.h
```diff
@@ -7,7 +7,7 @@
 
 #if defined(__aarch64__)
 #include "arch/arm/rte_atomic_64.h"
-else
+#else
```
mynktl, line 10: why are we removing this?

shubham14bajpai: Not removing, just added `#` which was missing before the `else`.

mittachaitu, line 10: :+1:

build_image.sh
shubham14bajpai: Done

mynktl (approved):

mittachaitu (approved):
changes are good

kmova (approved):

PR #311, opened 2020-03-04
fix(resize): add resize test cases when replicas are in different states

mittachaitu:
This PR adds different test cases to cover resize functionality.

Following are the test cases:
- When replicas are not connected trigger resize and expect iscsi disk of the same size after replicas are connected(offline resize).
- When all the replicas are in degraded state triggering resize.
- When only two replicas are connected out of three replicas then triggering resize.
- When only one replica is connected out of three replicas then triggering resize.

### DIFF
--- a/test_istgt.sh
+++ b/test_istgt.sh
@@ -239,13 +239,13 @@ write_data()
 setup_test_env() {
 	rm -f /tmp/test_vol*
 	mkdir -p /mnt/store
-	## We are doing resize to 7G
-	truncate -s 7G /tmp/test_vol1 /tmp/test_vol2 /tmp/test_vol3 $1
+	## We are doing multiple resizes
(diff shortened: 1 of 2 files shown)


test_istgt.sh
```diff
@@ -985,6 +985,141 @@ run_replica_connection_test ()
...
+	DESIRED_REPLICATION_FACTOR=3
+	REPLICATION_FACTOR=3
+	CONSISTENCY_FACTOR=2
+	VOLUME_SIZE=1G
+
+	setup_test_env
+
+	local vol_name=$(cat $CONF_FILE | grep TargetName | awk '{print $2}')
+	local new_size=2
+	echo "Resize of volume when replicas are in different state"
```
mynktl: There won't be any replica connected at this stage.

mittachaitu: I will update the message

vishnuitta (approved):
changes are good

mynktl (commented):
lgtm :+1:

PR #317, opened 2020-03-12
fix(image): update istgt dockerfile to use ubuntu 18.04

prateekpandey14:
- commit update the Dockerfile to use ubuntu 18.04 as a base
image to fix some of the security vulnerabilities with 16.04
image used earlier.
- Update the travis build to use bionic (18.04) 
- Change ioping command with count

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: c
 sudo: required
-dist: trusty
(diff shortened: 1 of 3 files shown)


vishnuitta (approved):
changes are good

PR #324, opened 2020-04-20
feat(env): IO_MAX_WAIT_TIME setting for replica timeout

vishnuitta:
This PR adds an environment variable "IO_MAX_WAIT_TIME" for istgt.
This setting is the maximum time istgt need to wait for response from replica. If replica doesn't respond with in this configured value, istgt will disconnect that replica.

### DIFF
--- a/src/init.sh
+++ b/src/init.sh
@@ -76,9 +76,9 @@ service rsyslog start
 
 if [ -z $test_env ]
 then
-	/usr/local/bin/istgt -R 20
+	IO_MAX_WAIT_TIME=120 /usr/local/bin/istgt -R 20
 else
-	QueueDepth=5 Luworkers=9 /usr/local/bin/istgt -R 20
+	IO_MAX_WAIT_TIME=120 QueueDepth=5 Luworkers=9 /usr/local/bin/istgt -R 20
 fi
 
 child=$!
--- a/src/replication.c
+++ b/src/replication.c
@@ -2009,6 +2009,7 @@ int istgt_lu_destroy_snapshot(spec_t *spec, char *snapname)
 		send_replica_snapshot(spec, replica, 0, snapname, ZVOL_OPCODE_SNAP_DESTROY, NULL);
 	TAILQ_FOREACH(replica, &spec->non_quorum_rq, r_non_quorum_next)
 		send_replica_snapshot(spec, replica, 0, snapname, ZVOL_OPCODE_SNAP_DESTROY, NULL);
+	REPLICA_LOG("snap %s destroy\n", snapname);
 	return true;
 }
 
@@ -2149,8 +2150,8 @@ int istgt_lu_create_snapshot(spec_t *spec, char *snapname, int io_wait_time, int
 	spec->quiesce = 0;
 	MTX_UNLOCK(&spec->rq_mtx);
 
-	REPLICA_LOG("snap create ioseq: %lu resp: %s\n", io_seq,
-	    (r == true) ? "success" : "failed");
+	REPLICA_LOG("snap %s create ioseq: %lu resp: %s successcnt: %d\n", snapname, io_seq,
+	    (r == true) ? "success" : "failed/timedout", success);
 	return (ret);
 }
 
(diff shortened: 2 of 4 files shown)


test_istgt.sh
```diff
@@ -2006,6 +2006,20 @@ run_io_timeout_test()
 		# Test to verify disconnection of replica if delay from replica is more than maxiowait
 		$ISTGTCONTROL maxiowait 5
 		$IOPING  -c 3 -B -WWW /dev/$device_name > $iopinglog
+		cmd="grep -ch 'changing IO max wait time from 60 to 120' $LOGFILE"
+		val=$(eval $cmd)
+		if [ $val -ne 1 ]
```
mynktl: count won't match because we are not cleaning the `/tmp/istgt.log`. Better to update `stop_istgt` with `rm $LOGFILE`.

mynktl (changes requested):
changes looks good. one fix required for test_istgt.tsh

test_istgt.sh
vishnuitta: fixed it by checking for 'eq 0' for error

mynktl (approved):
looks good to me :+1:  
Thanks @vishnuitta !

src/replication.c
```diff
@@ -4348,6 +4349,17 @@ initialize_replication()
 		return -1;
 	}
 	clock_gettime(CLOCK_MONOTONIC_COARSE, &istgt_start_time);
+	const char *s_max_wait_time = getenv("IO_MAX_WAIT_TIME");
+	int max_wait_time= 0;
+	if (s_max_wait_time != NULL)
+		max_wait_time = (int)strtol(s_max_wait_time,
+		    NULL, 10);
+	if (max_wait_time > 30) {
```
mittachaitu, line 4357: So with this configuration, we are allowing users to decrease the response time value from 60 to 31. Is this expected to happen?

src/init.sh
```diff
@@ -76,9 +76,9 @@ service rsyslog start
 
 if [ -z $test_env ]
 then
-	/usr/local/bin/istgt -R 20
+	IO_MAX_WAIT_TIME=120 /usr/local/bin/istgt -R 20
```
mittachaitu, line 79: Why are we setting `IO_MAX_WAIT_TIME=120` in both cases?

mittachaitu (commented):
Changes are good but need clarification on test case

test_istgt.sh
```diff
@@ -2006,6 +2006,18 @@ run_io_timeout_test()
 		# Test to verify disconnection of replica if delay from replica is more than maxiowait
 		$ISTGTCONTROL maxiowait 5
 		$IOPING  -c 3 -B -WWW /dev/$device_name > $iopinglog
+		cmd="grep -ch 'changing IO max wait time from 60 to 120' $LOGFILE"
```
mittachaitu, line 2009: Since this test is related to verifying env. Can we push related lines of code to `run_test_env` function so that we can avoid setting `IO_MAX_WAIT_TIME` in all the cases?

src/replication.c
vishnuitta: if user wants, it can happen

src/init.sh
vishnuitta: both the places we are setting the env.. requirement is to set the env, not the value

test_istgt.sh
vishnuitta: this test is about verifying env, and also make sure that 'istgtcontrol maxiowait' changes the value

mittachaitu (approved):
changes are good

mynktl:
@vishnuitta  let's add a changelog file, named `changelogs/unreleased/324-vishnuitta`,  with one-linear description of the change

vishnuitta:
> @vishnuitta let's add a changelog file, named `changelogs/unreleased/324-vishnuitta`, with one-linear description of the change

I'm adding a oneliner as `feat(env): IO_MAX_WAIT_TIME env setting for replica timeout`

mynktl (approved):

PR #333, opened 2020-05-19
chore(build): enhance image size

mittachaitu:
This PR enhances the size of the cstor-istgt image from 615MB to 272MB.
```sh
k8s@e2e1-master:~/Desktop/sai/istgt$ sudo docker images
REPOSITORY                           TAG                 IMAGE ID            CREATED             SIZE
openebs/cstor-istgt                    ci                  634d5d645390        About an hour ago   272MB
quay.io/openebs/cstor-istgt          ci                  fab9e11921ec        2 days ago          615MB
```

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:
- Followed the best practice mentioned [here](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/).

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Labelled this PR & related issue with `documentation` tag
- [ ] PR messages has document related information
- [ ] Labelled this PR & related issue with `breaking-changes` tag
- [ ] PR messages has breaking changes related information
- [ ] Labelled this PR & related issue with `requires-upgrade` tag
- [ ] PR messages has upgrade related information
- [ ] Commit has unit tests
- [ ] Commit has integration tests


build_image.sh
```diff
@@ -32,7 +32,7 @@ else
 	exit 1
 fi
 make clean
-make
+make -j4
```
kmova: q_: What does this mean?

mittachaitu: It will launch 4 jobs to run simultaneously which will help to compile the code quickly.

mittachaitu: I observed quick compilation using `make -j4` than just `make`.

docker/Dockerfile
```diff
@@ -3,11 +3,23 @@
...
+    net-tools \
+    iputils-ping \
+    gdb \
+    apt-utils \
+    libssl-dev \
+    libjson-c-dev \
+    libjemalloc-dev \
+    apt-file && apt-file update \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
```
mynktl, line 22: Can we move this command to new RUN since this is about cleaning the apt data.

mynktl: Same for arm64 docker file.

mynktl (approved):

docker/Dockerfile
mittachaitu: If we move to the next layer then it will not impact image size as mentioned in best practices.

> In addition, when you clean up the apt cache by removing /var/lib/apt/lists it reduces the image size, since the apt cache is not stored in a layer. Since the RUN statement starts with apt-get update, the package cache is always refreshed prior to apt-get install.

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

build_image.sh
pawanpraka1: @mittachaitu please change this to make -j$(nproc).

pawanpraka1 (changes requested):
use make -j$(nproc).

build_image.sh
mittachaitu: Updated PTAL

pawanpraka1 (approved):
looks good.

docker/Dockerfile
mynktl: Yes. My bad. It won't remove files from older layer unless you use `--squash` and  that is experimental feature, not supported by default.

mynktl (approved):

kmova (approved):

PR #338, opened 2020-05-29
refact(release): add cstor-operators in downstream

kmova:

The current release script(in travis) only tags
the external-storage and maya, that have control
plane operators for SPC.

The new control plane oprators for CSPC are under
cstor-operators repo.

**Which issue this PR fixes**:

After building the cstor-istgt containers,
tag cstor-operators to build and release the
new control plane operators.

Signed-off-by: kmova <kiran.mova@mayadata.io>


mittachaitu (approved):
Changes are good(So, flow is cstor build will trigger istgt then istgt will trigger cstor-operaters build).

PR #354, opened 2021-04-09
chore(actions): move travis tests to github action

shubham14bajpai:
Signed-off-by: shubham <shubham.bajpai@mayadata.io>

This PR moves the travis test to github actions. This PR also fixes the build and release workflow to get executed according to the events. 

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Labelled this PR & related issue with `documentation` tag
- [ ] PR messages has document related information
- [ ] Labelled this PR & related issue with `breaking-changes` tag
- [ ] PR messages has breaking changes related information
- [ ] Labelled this PR & related issue with `requires-upgrade` tag
- [ ] PR messages has upgrade related information
- [ ] Commit has unit tests
- [ ] Commit has integration tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,10 +13,20 @@
 # limitations under the License.
 name: build
 
-on: ['push']
+on:
+  create:
+  push:
+    branches:
+    - replication
+    - 'v*'
+    paths-ignore:
+    - 'docs/**'
+    - 'changelogs/**'
+    - '*.md'
 
 jobs:
   lint:
+    if: ${{ (github.event.ref_type != 'tag') }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
@@ -30,12 +40,33 @@ jobs:
           path: '.'
           pattern: '*.sh'
 
-  istgt:
-    runs-on: ubuntu-latest
+  tests:
+    if: ${{ (github.event.ref_type != 'tag') }}
     needs: ['lint']
+    runs-on: ubuntu-18.04
+    timeout-minutes: 120
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      
+      - name: Build
+        run: ./ci/build.sh
+
+      - name: Running tests
+        run: ./ci/test.sh
+
+  istgt:
+    runs-on: ubuntu-latest
+    needs: ['tests']
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set Image Org
+        # sets the default IMAGE_ORG to openebs
+        run: |
+          [ -z "${{ secrets.IMAGE_ORG }}" ] && IMAGE_ORG=openebs || IMAGE_ORG=${{ secrets.IMAGE_ORG}}
+          echo "IMAGE_ORG=${IMAGE_ORG}" >> $GITHUB_ENV
 
       - name: Set tag
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -30,13 +40,34 @@ jobs:
           path: '.'
           pattern: '*.sh'
 
+  tests:
+    if: ${{ (github.event.ref_type != 'tag') }}
```
akhilerm: If the test requires `lint`, this check is not needed.

.github/workflows/build.yml
```diff
@@ -69,9 +118,23 @@ jobs:
...
+          context: .
+          file: ./docker/istgt.Dockerfile
+          push: true
+          platforms: linux/amd64, linux/arm64
+          tags: |
+            ${{ steps.docker_meta.outputs.tags }}
+          build-args: |
+            DBUILD_DATE=${{ steps.date.outputs.DATE }}
+            DBUILD_REPO_URL=https://github.com/openebs/istgt
+            DBUILD_SITE_URL=https://openebs.io
```
akhilerm, line 139: Doesn't this require the `BRANCH` env ?

.github/workflows/build.yml
```diff
@@ -30,13 +40,34 @@ jobs:
...
+        uses: actions/checkout@v2
+      
+      - name: Build
+        run: ./ci/build.sh
+
+      - name: Running tests
+        run: ./ci/test.sh
+
   istgt:
     runs-on: ubuntu-latest
```
akhilerm, line 58: The tests and actual build runs on 2 types of VMs. One is `ubuntu-latest` and the other is `ubuntu-18.04`. May need to fix on using one version

.github/workflows/pull_request.yml
```diff
@@ -54,7 +72,12 @@ jobs:
         with:
           version: v0.5.1
 
-      - name: Build Image
-        env:
-          IMG_RESULT: cache
-        run: make -f Makefile.buildx.mk docker.buildx.istgt
+      - name: Build & Push Image
```
akhilerm:
```suggestion
      - name: Build Image
```

ci/build.sh
```diff
@@ -0,0 +1,46 @@
...
+# related code doesn't break the existing target code.
+./configure CC="gcc-6" CXX="gcc-6"
+make
+make clean
+ARCH=$(uname -m)
+echo ARCH "$ARCH"
+if [ "${ARCH}" = "x86_64" ]; then
+    ./configure CC="gcc-6" CXX="gcc-6" --enable-debug --enable-replication
+elif [ "${ARCH}" = "aarch64" ]; then
+    ./configure CC="gcc-6" CXX="gcc-6" --enable-debug --enable-replication --build=arm-linux
```
akhilerm, line 43: Is this same file also called while using the buildx build? Because if arch specific rules are there, we may need to check that.

test_istgt.sh
```diff
@@ -454,15 +454,17 @@ run_read_consistency_test ()
 		return
 	fi
 
-	write_data 0 41943040 512 "/dev/$device_name" $file_name
+	write_data 0 41943040 4096 "/dev/$device_name" $file_name
```
akhilerm, line 457: Any reason why the block size was changed?

akhilerm (commented):
@shubham14bajpai Given some comments

.github/workflows/build.yml
shubham14bajpai: No as no goflags are not required here

.github/workflows/build.yml
shubham14bajpai: Here in tests the code is build locally so it depends on the os, whereas in the istgt the code is build in a docker container build stage which has ubuntu-18 as the build stage image. https://github.com/openebs/istgt/blob/3a853b354eab46258f46331bdea4bdf4c7e549ab/docker/istgt.Dockerfile#L15

ci/build.sh
shubham14bajpai: Yes but some flags are not required in the final build and only test so we have a separate file 
https://github.com/openebs/istgt/blob/replication/docker/build.sh

test_istgt.sh
shubham14bajpai: The consistency test was failing as the block size was different in two commands. cc @mynktl

ci/build.sh
akhilerm: Can these both files be combined, or will it add confusion?

shubham14bajpai: It will make the script and action more complex as we need to add envs and nested ifs to decide whether to build with debug flags or not. It is better to keep them separate.

mynktl (approved):

test_istgt.sh
akhilerm: @mynktl  Can you explain on these. How did the tests passed earlier?

akhilerm (approved):
LGTM. 

Just a comment on the block size. cc: @mynktl

test_istgt.sh
mynktl: In this particular test, we are duplicating stream/writes to the iscsi volume and local file, and checking if both are having the same data or not.

Initially, we are writing ~40Mb of data to device and local files using 512 blocksize. and in the second stage, we are writing ~10Mb using 4096 blocksize. In both stages, we are writing from offset 0.

Writing 40Mb of data will result in an error due to block size mismatch. Although some initial writes may have succeeded for a local file in the first stage, the second write of 10Mb will make sure that both have the same data.

kmova (approved):

